PR #1, opened 2018-02-20
Base structure

wtrocki:
## Motivation 

Provide base structure that could be used no matter what base language we will pick.
Avoid controversial changes and bring all the goodness every repository needs.

Changes include:

- Github helper files
- Contributor guide
- Documentation templates 
- Initial folder structure that will be extended in separate PR,
I wanted to flag how structure should look like
- Placeholders for example applications (opening discussion if we need them etc.)
- Enable build for circleci

Notes: I intentionally did not pushed core implementation as it may be different depending on language/technology. Going to push separate PR once this will be merged.
Going to split the initial work into separate PR's to simplify reviews.


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
```
secondsun: http://editorconfig.org/ if you are a dummy like me :)

README.md
```diff
@@ -1,2 +1,44 @@
-# aerogear-web-sdk
+# AeroGear Web SDK
+[![License](https://img.shields.io/badge/-Apache%202.0-blue.svg)](https://opensource.org/s/Apache-2.0)
```
secondsun: Can you add a circle ci build status shield?

wtrocki: Good point! Done

wtrocki:
It took less than 30 seconds to run install and tests on CircleCi. I love that platform!

PR #2, opened 2018-02-20
Base scripts and core in TypeScript

wtrocki:
## Motivation

Implementation for Core in TypeScript with all required build systems and lerna integration.
This PR was used initially to compare JS and TypeScript aproach.


package.json
```diff
@@ -1,12 +1,30 @@
 {
-	"name": "AeroGearServicesWeb",
+	"name": "raincatcher",
```
david-martin: copy/paste error?

wtrocki: I do not care as much as this code may be reverted. I wanted to get some decision made on typescript vs plain JS vs Webpack  and have it close to PR.

tsconfig.json
```diff
@@ -0,0 +1,58 @@
...
+    // "noUnusedParameters": true,            /* Report errors on unused parameters. */
+    // "noImplicitReturns": true,             /* Report error when not all code paths in function return a value. */
+    // "noFallthroughCasesInSwitch": true,    /* Report errors for fallthrough cases in switch statement. */
+
+    /* Module Resolution Options */
+    // "moduleResolution": "node",            /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
+    // "baseUrl": "./",                       /* Base directory to resolve non-absolute module names. */
+    // "paths": {},                           /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */
+    // "rootDirs": [],                        /* List of root folders whose combined content represents the structure of the project at runtime. */
+    // "typeRoots": [],                       /* List of folders to include type definitions from. */
```
pb82: should probably be set to `node_modules/@types`

pb82:
@wtrocki thanks, i think this branch can be used as the TS example for the discussion around JS vs. TS

tsconfig.json
wtrocki: That will be default. This repo will use hoisting, so also root types will match.
Default option includes that so I'm not going to do it.

wtrocki:
@pb82 @david-martin Ready for the final review. Things are now typescripty and everything should be clean.
I have more code/examples to be added as separate PR. We just need base on master to unblock react native.

package.json
```diff
@@ -4,9 +4,27 @@
 	"description": "AeroGear Services Web SDK package",
 	"main": "index.js",
 	"scripts": {
-		"test": "lerna run test"
+		"test": "lerna run test",
+		"test-ci": "npm-run-all test coverage:*",
```
aliok: This one failed locally for me. Just a FYI, no big deal.

packages/core/src/Config.ts
```diff
@@ -0,0 +1,52 @@
+/**
```
aliok: any reason to have this file's name starting with capital letter?

packages/core/test/Config.ts
```diff
@@ -0,0 +1,27 @@
+import * as  chai from 'chai';
+import * as sinon from 'sinon';
+import { ConfigService } from '../src/Config';
+
+/// <reference types="node" />
```
aliok: Are these "reference" lines still a thing :( ? The old version I still use privately has these (or tsconfig.json, which is my preference) but I was hoping that they got rid of these in the future versions.

paolobueno:
The addition of all the tooling scripts is very important but I think the example for Config.ts + tests isn't already enough to shed much light into the mail thread discussion around the language for this SDK.

Follow up PRs should be even more interesting :+1:

packages/core/.gitignore
```diff
@@ -0,0 +1,6 @@
+# generated code
+src/**/*.js
```
paolobueno: In retrospect (from raincatcher) I'd rather output the js to `dist/` and have a watcher->recompile command. No need for tacking this now of course.

packages/core/src/Config.ts
paolobueno: It should be because it exports the ConfigService class as default.

@wtrocki if you had concerns about usage from JS code this could be lowercase `config.ts` and avoid using the default export, that way JS could still do

```javascript
let {IServiceConfig, IAeroGearConfig, ConfigService} = require("./config");
```

packages/core/test/Config.ts
paolobueno: I don't think this is necessary if `@types/node` is installed, not sure off the top of my head.

package.json
wtrocki: Great catch!

packages/core/src/Config.ts
wtrocki: @paolobueno - this could be good style example. In RainCatcher we made style for Uppercase letters. 
I will avoid default export - default exports are cancer :P

packages/core/test/Config.ts
wtrocki: Code worked but visual studio complained. 
This is only to improve VisualStudio experience.

packages/core/.gitignore
wtrocki: Purpose of this PR is mostly to tackcle infrastructure/scripts/build for typecripts.
More business specific code will be added in other PR.

I got impression that dist will make unit testing and publishing more complex.

packages/core/src/Config.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Service configuration model
+ */
+export interface IServiceConfig {
```
pb82: Not sure about the `I` prefix for interfaces. Microsoft seems to discourage this: https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines (under `Names`)

packages/core/.gitignore
paolobueno: @wtrocki yes, I think there are always tradeoffs.
For using TS+lots of lerna modules we decided to output the compiled files to the same directories and include TS sources in the final npm packages instead of generated typings. This speeds up SDK development due to faster navigation and no need to constantly recompile internal dependencies.

However this also means that TS clients were trying compile from TS sources when getting the packages from npm, and that led to confusion sometimes.

An alternative is setting up file watchers for the packages and recompiling to js in a different directory like `dist/` and having `package.json#main = "./dist/index.js"`

It's a tradeoff between convenience during SDK development vs having the dev environment closer to the final published package shape... I don't think we need to worry about this right now though.

wtrocki:
@paolobueno Going to create follow up PR together with web story (webpack)

PR #3, opened 2018-02-21
initial react-native example

pb82:
## Motivation

Add an initial react-native example. Simple Navigation included with only one Welcome page. No SDK usage at the moment because we don't yet have any JS SDKs.

Includes a README to help users get started.
Uses the AeroGear Logo as the App icon.


wtrocki:
> I would add some Redux boilerplate and some eslint rules.

That is vanilla example - it's not intended to build on. 
It should be minimal as possible.

example/react-native/android/app/build.gradle
```diff
@@ -0,0 +1,150 @@
+apply plugin: "com.android.application"
```
wtrocki: Not sure if we can avoid pushing actual platform source code?

josemigallas:
@wtrocki The files here are the minimum scaffold for a real React Native app. The example you are referencing is generated by [create-react-native-app](https://github.com/react-community/create-react-native-app) which is only meant to be using [Expo](https://expo.io/). I believe this project was created using `react-native init`.

pb82:
@wtrocki @josemigallas yes, the boilerplate was generated using `react-native init`. I would prefer to leave the platform code in the repo. As far as i can see, other popular projects (e.g. https://github.com/iSimar/HackerNews-React-Native, https://github.com/7kfpun/FinanceReactNative, ...) do this too. Otherwise the project would not work with `react-native run-<platform>`

pb82:
@josemigallas Maybe we can add Redux to the example when we have a use case with the SDK where it makes sense (auth).

josemigallas:
Yes, that was just my suggestion, I think this is good already

PR #7, opened 2018-02-22
Temporary contributing guide

wtrocki:
## Motivation

Contributing guide (developer getting started guide). 
Guide will evolve over the time. I wanted to provide just minimal set of instructions and general idea how SDK will be used.


david-martin:
@wtrocki I'll give this a spin

PR #10, opened 2018-02-22
Fix running of unit tests

paolobueno:
@wtrocki This should allow for running unit tests

### DIFF
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -4,8 +4,7 @@
     "src/"
   ],
   "compilerOptions": {
-    "rootDir": "./src",
     "outDir": "./dist/",
     "declarationDir": "./types"
-  },
+  }
 }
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -6,7 +6,6 @@
     "strictNullChecks": true,
     "esModuleInterop": true,
     "noImplicitAny": true,
-    "module": "es6",
```
paolobueno: Had to remove this because ts-node was compiling ts to es6 modules instead of commonjs and mocha running on node couldn't handle them.

packages/core/tsconfig.json
```diff
@@ -4,8 +4,7 @@
     "src/"
   ],
   "compilerOptions": {
-    "rootDir": "./src",
```
paolobueno: This would make ts-node+tsc complain that the `test/` folder wasn't inside the root. I think the `"include"` array is enough.

PR #12, opened 2018-02-22
✨ Add the Core SDK to the react-native example

david-martin:
* Updated instructions to install the `react-native-cli`
* Core SDK gets installed with `npm install` from the packages folder
* Sample mobile-services.json file added (contents of keycloak config is
shown in example app)

### DIFF
--- a/example/react-native/README.md
+++ b/example/react-native/README.md
@@ -8,17 +8,24 @@ This is a small example app to demonstrate the features of the AeroGear SDK for 
 Make sure that npm and the react native CLI are installed on your system:
 
 ```
-$ brew install npm
-$ npm install -g react-native
+brew install npm
+npm install -g react-native-cli
 ```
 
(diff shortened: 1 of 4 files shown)


example/react-native/README.md
```diff
@@ -8,17 +8,24 @@ This is a small example app to demonstrate the features of the AeroGear SDK for
...
 
+Make sure you have the 'core' package built first. See [../../docs/contributing-guide.adoc](../../docs/contributing-guide.adoc)
+
 You can try the example by running:
 
 ```
-$ npm install
-$ react-native run-<platform>
+# Install dependencies
+# NOTE: This will also install a copy of the core sdk from ../../packages/core
```
wtrocki: I will try to automate that so all packages will be automatically linked by example.
Have couple ideas how we can use lerna for this particular use case.

PR #14, opened 2018-02-23
Umd modules

pb82:
## Motivation

WIP: Remove the `browser` and `module` properties from package.json and use UMD modules instead (https://github.com/umdjs/umd). react-native is not compatible with those properties because `browser` will override `main`. That leads to rn picking up a module packaged in a way it's not compatible with.

This change should work with all platforms but currently only react native is tested. Cordova does not yet make use of the core module, but i will test this before merging the PR.

This should fix @danielpassos 's issue with running the example.


.gitignore
```diff
@@ -73,3 +73,8 @@ package-lock.json
 ## Dist folders
 dist
 types
+
+# Module compiled output
+packages/core/src/*.js*
```
wtrocki: I do not get that compiled on the branch.

pb82: Hmm, i had these after running `npm run build`

paolobueno: I think this might come from `tsc` in `packages/core`, maybe https://github.com/aerogear/aerogear-js-sdk/blob/master/packages/core/package.json#L13 should be `webpack` ?

tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "esModuleInterop": true,
     "noImplicitAny": true,
     "target": "es5",
+    "module": "UMD",
```
wtrocki: 👍 for that. It was even suggested to do in previous PR.

wtrocki:
@pb82 If react native will work with UMD modules, cordova should as well.
We could even assume that we do not need to have self contained web delivery as people will almost always use packager.  With that in mind we do not need probably webpack at all and we could just figure out to recompile TypeScript on change using something like: https://www.npmjs.com/package/tsc-watch

Webpack is now slightly broken

```
WARNING in ./src/Config.ts
3:24-31 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
 @ ./src/Config.ts
 @ ./src/index.ts
```

@pb82 @paolobueno I think we getting really close that we need. Can you share what is your point on this topic?

pb82:
@wtrocki +100 for getting rid of Webpack. The README still says we should use `npm run build` instead of `npm run build:webpack`. I'll add SDK usage to the Cordova example and @danielpassos is currently also verifying the change on react native.

wtrocki:
@paolobueno - Is that something you would go as well?
@pb82 - Can I commit on top of this PR in order to make this changes visible?

pb82:
@wtrocki yes, you can try it. If it doesn't work then i can move this branch to origin.

paolobueno:
+1 Yes, I believe we could leave webpack and any further optimizations it would bring for projects that use the sdk to implement them as needed.

To be clear: just do regular `tsc` for final output, `tsc-watch` for local sdk dev.

wtrocki:
@pb82 Work done, but I cannot approve myself.

Note: this will mean that we do not support classical `<script src=""` use cases and our packages will always require browserify, webpack etc.
I just want to make sure that we are aware of this.

wtrocki:
For the cordova usage I think we should use scripts in actual demo app to link core (and other future SDK's) and browserify.
```
cd ../../packages/core/
npm link 
cd - 
npm link @aerogearservices/core
browserify index.js
```

See: https://github.com/aerogear/aerogear-js-sdk/blob/master/docs/contributing-guide.adoc

paolobueno:
UMD should work in a browser environment without extra tooling, we just need to check what name the modules are getting exported as.

By default they'll be added as globals, i.e. to `window['module-name']`, but if core currently is being set as `window.index` and/or `window.Config` then it wouldn't be easy to use.

wtrocki:
@Paolo I meant that people will need to use the tooling to bundle out SDK. UMD will work, but it will mean including every single js file in our SDK. Am I missing something? Lets not merge this as I'm confused or maybe not getting how this will work.


@irdavem Do you still have gist that worked for cordova?

pb82:
@wtrocki I think browserify + cordova could be as simple as `browserify www/js/index.js -o www/js/bundle.js` if we expect users to have it globally installed. We can also ship `bundle.js` in the example to make it work.

pb82:
@wtrocki @paolobueno @david-martin this is an example for how the SDK could be used in Cordova through browserify: https://github.com/aerogear/aerogear-js-sdk/compare/master...pb82:cordova-browserify?expand=1

(requires a globally installed browserify, but we could also include it in the dependencies)

wtrocki:
To make things clear. Since we will require people to bundle things for cordova and web does UMD format makes sensie. It feels like it generates a lots of boilerplate code for each file which is great because it will suppport we but we will not use this functionality. Can we just revert to commonjs format? Any opinions against?

paolobueno:
@wtrocki ah, I see what you mean that users would need many many `script` tags

@pb82 the idea behind having the "browser" config was that browser-targetting tools would pick it up automatically, I think in the end the react-native cli picking it up is just unfortunate.
Even removing the config we could still include the bundle so people can use `<script src="./node_modules/@aerogearservices/core/bundle.js">`

Thinking about the published npm package layout it would be something like:

```
- package.json: {main: dist/index.js}
...
- dist
  - bundle.js (umd that exposes a aerogearservices-core variable/module)
  - index.js (commonjs)
  - index.js.map
  - bundle.js.map(?)
  - types
    - *.d.ts
```

Just haven't found documentation about specifying the umd module name yet.

wtrocki:
@paolobueno That was the idea behind the webpack. Let have discussion about that to avoid removing/adding this in future. I think we are really close to make final call on how this should work.

@pb82 I'm glad and you stepped in to resolve this issues. Really appreciate that. 
This PR contains all possible configurations now. I will try to test it and post some thoughts on the list - to have wider scope.

wtrocki:
Made small change to fix breaking TypeScript usage.
From my point of view this change is ok to be merged, but wanted to make sure that we are aware of the scope of this. Since we no longer bundle our packages and web will require additional packager. With all this assumptions UMD format makes no sense as this files will be never consumed directly.

PR #15, opened 2018-02-23
Metrics service

wtrocki:
## Motivation

Proposal to open discussion around platform wrappers. 
I have used metrics engine implementation to work on the real case scenario that will be used in the near future.

## Main focus 

Open discussion what metrics make sense and how we collect them.
Showcase how metrics can be done for other platforms.

## Progress
- [x] Interfaces
- [x] Metrics
- [x] Engine
- [ ] Unit tests
- [ ] Documentation

### DIFF
--- a/packages/core/src/metrics/MetricsPublisher.ts
+++ b/packages/core/src/metrics/MetricsPublisher.ts
@@ -1,0 +1,34 @@
+import axios from 'axios';
+
+/**
+ * Interface for classes that can publish or store metrics payload
+ */
+export interface MetricsPublisher {
+  /**
+   * Allows to publish metrics to external source
+   */
+   publish(payload: any): void;
+}
+
+/**
+ * Metrics publisher that sends payload to remote server
+ * Publisher requires remote server URL
+ */
+export class MetricsNetworkPublisher implements MetricsPublisher {
+
+  constructor(private url: string) {
+  }
+
+  /**
+   * Allows to publish metrics to external source
+   */
+  public publish(payload: any): void {
+    axios.post(this.url, payload)
+    .then(function(response) {
+      console.info(response);
+    })
+    .catch(function(error) {
+      console.error(error);
+    });
+  }
+}
--- a/packages/core/src/metrics/impl/AppMetrics.ts
+++ b/packages/core/src/metrics/impl/AppMetrics.ts
@@ -1,0 +1,21 @@
+import {Metrics} from '../Metrics';
+
+/**
+ * Collect application metrics:
+ *
+ *  - appId - application bundle id
+ *  - appVersion - version of the application
+ *  - sdkVersion - AeroGear Services SDK version
+ */
+export class AppMetrics implements Metrics {
+  public identifier: string = 'app';
+  public collect(): any {
+    // TODO - does this even make sense to collect
+    const data: any = {
+      'appId': 1,
+      'appVersion': 1,
+      'sdkVersion': 1
+    };
+    return data;
+  }
+}
--- a/packages/core/src/metrics/impl/DeviceMetrics.ts
+++ b/packages/core/src/metrics/impl/DeviceMetrics.ts
@@ -1,0 +1,19 @@
+ import { Metrics } from '../Metrics';
+
+/**
+ * Collect device metrics:
+ *
+ *  - platform - ios
+ *  - platformVersion - version of the ios platform
+ *  - device - device name
+ */
+ export class DeviceMetrics implements Metrics {
+  public identifier: string = 'device';
+  public collect(): any {
+    return {
+      'platform': 'js',
+      'platformVersion': 1,
+      'device': 'TODO - cordova/react wrapper'
+    };
+  }
+ }
--- a/packages/core/src/wrapper/DeviceData.ts
+++ b/packages/core/src/wrapper/DeviceData.ts
@@ -1,0 +1,21 @@
+import { isCordovaApp, isReactNative } from './PlatformHelpers';
+
+// [Investigation] Detecting platforms on runtime.
+
+// Welcome to TypeScript hacks
+declare var device: any;
+// Welcome to TypeScript hacks
+declare var Platform: any;
+
+/**
+ * Get device platform
+ */
+export function getPlatform(): string {
+    if (isCordovaApp) {
+      // https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-device/
+      return device.platform;
+    } else if (isReactNative) {
+      return Platform.OS;
+    }
+    return 'Unknown';
+}
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -34,6 +34,7 @@
     "branches": 80
   },
   "devDependencies": {
+    "@types/axios": "^0.14.0",
     "@types/chai": "4.1.2",
     "@types/chai-as-promised": "7.1.0",
     "@types/cordova-plugin-camera": "2.4.0",
@@ -53,7 +54,6 @@
     "webpack": "^3.11.0"
   },
   "dependencies": {
-    "@types/lodash": "4.14.104",
-    "lodash": "4.17.5"
+    "axios": "^0.18.0"
   }
 }
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -3,3 +3,4 @@
  */
 
 export * from './Config';
+export * from './metrics/MetricsService';
--- a/packages/core/src/metrics/Metrics.ts
+++ b/packages/core/src/metrics/Metrics.ts
@@ -1,0 +1,47 @@
+import {MetricsPublisher} from './MetricsPublisher';
+
+/**
+ * Interface used for mobile metrics management
+ * Allows other SDK and implementations to manage (add) metrics
+ */
+export interface MetricsPublishable {
+
+  /**
+   * Allows to override default metrics publisher
+   *
+   * @param publisher - implementation of metrics publisher
+   */
+  setMetricsPublisher(publisher: MetricsPublisher): void;
+
+  /**
+   * Collect application and device metrics
+   * Send data instantly using active metrics publisher
+   */
+  sendAppAndDeviceMetrics(): void;
+
+  /**
+   * Publish user defined metrics
+   *
+   * @param metrics Varargs of objects implementing MetricsCollectable
+   */
+  publish(...metrics: Metrics[]): void;
+}
+
+/**
+ * Interface used for for classes that will colect metrics
+ */
+export interface Metrics {
+  /**
+   * A identifier that is used to namespace the metrics data
+   *
+   * @return identifier string
+   */
+  identifier: string;
+
+  /**
+   * Function called when metrics need to be collected
+   *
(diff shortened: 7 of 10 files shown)


packages/core/src/metrics/impl/DeviceMetrics.ts
```diff
@@ -0,0 +1,18 @@
...
+ *  - platformVersion - version of the ios platform
+ *  - device - device name
+ */
+ export class DeviceMetrics implements Metrics {
+  public identifier: string = 'device';
+  public collect(): MetricsData {
+    return new MetricsData([
+      'platform': 'js',
+      'platformVersion': 1,
+      'device': 'TODO - cordova wrapper'
```
wtrocki: @pb82 - this cold be actually really good use case for react native/cordova wrapper support.
Pinging you for awareness.

packages/core/src/metrics/impl/DeviceMetrics.ts
```diff
@@ -0,0 +1,20 @@
...
+ *  - platformVersion - version of the ios platform
+ *  - device - device name
+ */
+ export class DeviceMetrics implements Metrics {
+  public identifier: string = 'device';
+  public collect(): any {
+    return {
+      'platform': 'js',
+      'platformVersion': 1,
+      'device': 'TODO - cordova/react wrapper'
```
wtrocki: @pb82 @david-martin  this cold be actually really good use case for react native/cordova wrapper support.
Pinging you for awareness.

packages/core/src/metrics/impl/AppMetrics.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  - appId - application bundle id
+ *  - appVersion - version of the application
+ *  - sdkVersion - AeroGear Services SDK version
+ */
+export class AppMetrics implements Metrics {
+  public identifier: string = 'app';
+  public collect(): any {
+    // TODO - does this even make sense to collect
+    const data: any = {
+      'appId': 1,
```
wtrocki: This will make sense for cordova/react. There is not much sense to have this metrics for web.

packages/core/src/metrics/MetricsPublisher.ts
```diff
@@ -0,0 +1,34 @@
+import axios from 'axios';
```
david-martin: @wtrocki What is axios?
Is it some cross platform (cordova, react native & web) http(s) lib?
Is this a 'de-facto' lib for solving this kind of problem?
Is there no native http(s) thing (`fetch` api?) for doing this

wtrocki: @david-martin I have done research for the networking layers.

pb82: @david-martin correct, it's a cross platform http library. The standard is `fetch`, i think the only advantage of axios is only that is has more features and also works on Node.js (not relevant for us though).

david-martin:
@wtrocki It's great to have this WIP so soon.
However, I'd like to keep it open for some time until the iOS & Android metrics sdks are nailed down, and base any other platform/target off those.

Also, I think there's more value in figuring out how any native code/dependency would work & look for the JS SDK, which is part of what https://issues.jboss.org/browse/AEROGEAR-2064 covers (cc @pb82)
If we had a good example of how native code works for Cordova & React Native in a singular JS SDK, we'd be able to desrisk more.

wtrocki:
> Also, I think there's more value in figuring out how any native code/dependency would work & look for the JS SDK, which is part of what https://issues.jboss.org/browse/AEROGEAR-2064 covers (cc @pb82)

@david-martin Please ready my comment above where I pinged you and Peter. This PR is actually giving us *real* use case and scenarios where this can happen. 

> If we had a good example of how native code works for Cordova & React Native in a singular JS SDK, we'd be able to desrisk more.

This is what this PR is about. See use cases and separate PR that gives us alternative aproach. Please read PR content before making this kind of comments.

david-martin:
> This is what this PR is about. See use cases and separate PR that gives us alternative aproach.

Thanks @wtrocki. I must have missed something if that's the case.
Can you point me to where there is some call out to native code e.g. to a cordova plugin or react module. Something that absolutely requires native code (I'm thinking that would be something like security check for a rooted device, but that's specific to security, not metrics)

wtrocki:
It's literally above your comments:
screen shot 2018-02-26 at 10 32 19 am

pb82:
@wtrocki Platform/Platform version can be retrieved at least in react-native easily. Device not so much (but it's also not documented in the metrics proposal). App ID (package/module name) is difficult (impossible) without native code though. I'm currently writing an analysis for this.

wtrocki:
@pb82 - Found plugin for that in cordova that I'm using:
https://github.com/whiteoctober/cordova-plugin-app-version

Currently I have really dirty cordova example so I do not want to push it as it will hide real changes.

This PR is showcasing one way how we can architecture wrappers for different platforms, but there will be much more options that will give us more benefits.

pb82:
@wtrocki But how do we integrate those cordova plugins / react-native libs into our SDK so that the user can transparently require them?

david-martin:
> @wtrocki But how do we integrate those cordova plugins / react-native libs into our SDK so that the user can transparently require them?

@wtrocki This is what I'd really love to see.
This PR doesn't show that.
I'd really like to see how native code gets integrated into the SDK, and where the touchpoints are for cordova plugin code vs. react-native code.
Does that make sense?

packages/core/src/wrapper/DeviceData.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export function getPlatform(): string {
+    if (isCordovaApp) {
+      // https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-device/
+      return device.platform;
+    } else if (isReactNative) {
+      return Platform.OS;
+    }
+    return 'Unknown';
+}
```
wei-lee: -1 on this approach. It's not really reusable, and hard to extend. 

A better approach I think we can use is we should take advantage of the interfaces feature provided by TS, and define a common interface to provide the information that we need, and then provide different implementation backed by Cordova or ReactiveNative.

In this case, we can have:

```
interface Device {
  string getPlatform();
}

class CordovaDevice {
  string getPlatform() {
    return device.platform
  }
}

class ReactNativeDevice {
  string getPlatform() {
    return Platform.OS;
  }
}
```

Then during the init time, the core module will only need to detect the platform once, and then instantiate the right implementation automatically.

wtrocki:
> @wtrocki But how do we integrate those cordova plugins / react-native libs into our SDK so that the user can transparently require them?

We need to talk about it because I think that there is major misconception what is the challenge here. Wrapping native code will be just standard way of building cordova and react native plugins. The real challenge here will be to build mechanism/wrapper/architecture to be able to seamlessly switch between platform code. Some of the platforms may even require separate implementations etc. so we will need to make this wrapper so flexible that person working on the platform will not be restricted by limitations of the other. In terms of integration of the native plugins into app - this really well understood problem that you can google and there are a lots of proven ways to do it.

If native code is available as NPM modules for cordova and react native plugins then we

packages/core/src/wrapper/DeviceData.ts
wtrocki: Great! I think the same. I have done something like this alternative aproach: https://github.com/aerogear/aerogear-js-sdk/pull/16

What I think now is that we need something in the middle - we do not need separate module as in PR above, but we need to have wrappers and clasess as you have suggested.

packages/core/src/metrics/MetricsPublisher.ts
wtrocki: @pb82 Fetch is not supported seamlessly on web: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

david-martin:
> just standard way of building cordova and react native plugins

> really well understood problem

@wtrocki That's excellent that there is a standard way of building, and using plugins is really well understood.
However, I would love to see an example where these standards, and really well understood problem are brought together in a single SDK that works:
* in a seamless way for the mobile developer regardless of platform (I think that's covered quite well by this PR)
* in a way that allows the SDK developer (us) to package up the SDK, including any native dependencies, for easy inclusion in a project. This bit isn't clear, and would really benefit from a 'steel thread' example where there is a native dependency.

Some questions to answer as part of giving that example:
* do we bundle all native dependencies (for Cordova & React Native) in the SDK? Or are dependecies installed by the mobile developer?
* do we release a separate SDK for Cordova & React Native, both bundled from the 1 repo? Or is there a single JS SDK artifact/release that works for both?

I'm OK with an example being done in a separate, focused PR. (And it doesn't necessarily have to be @wtrocki who does it, given the task @pb82 is working on relates more to this)

@wtrocki Does this make sense?

packages/core/src/metrics/MetricsPublisher.ts
david-martin: @wtrocki Agreed.
Separate issue to figure out a 'cross platform' http lib.
That would depend on what range of devices/browsers we should support.
This may come in handy when doing that https://caniuse.com/#feat=fetch

wtrocki:
@david-martin - I'm worried that we went of the topic for this PR. Let's continue discussion on the email thread. Just going to present my point of view to not leave unanswered questions.

> do we bundle all native dependencies (for Cordova & React Native) in the SDK? Or are dependecies installed by the mobile developer?

That may not technically possible/challenging. It may have opposite effect to developer experience. Let's not really discuss that until we will make decision how packages are distributed. TL;DR - you would always follow the documentation of the Cordova and ReactNative without going against what they recommend. Typical aproach is that you working with plugins designed to do specific job. To visualize that imagine that we taking the Sync aproach. 

Sync documentation has just list of steps that are required for the platform.

1) For non wrapped api using existing plugins
```
cordova plugin add camera
cordova plugin add device
## Using config parser because I'm not sure if we need core package at all. 
npm install @aerogearservices/config-paser
```
Note that some of the plugins will be most likely there as they are added by default. 

2) If want to wrap and provide plugin (or plugin exist like push)

```
cordova plugin add push-server
## Using config parser because I'm not sure if we need core package at all. 
npm install @aerogearservices/config-paser
```

If we include this into the bundle we will make developers and our lives harder.
It will cause challenges on packaging. It could mess with build system, versioning, we will need to make this work with TypeScript which will mean spending countless hours on wrappers.

> 'steel thread' example where there is a native dependency.

Most of the cordova and react native examples will have this. 
I did not wanted to push that to this PR as I have really messy code on the example app.

> brought together in a single SDK that works

Wrapping all packages into some kind of uber bundle will be really bad for users. 
Ideally we could just have configuration parser working with some of the plugins or even example that will showcase that. I think that this will be better for Cordova and ReactNative community than providing uber bundle. This will allow us to be more aligned with the way cross platform development works and it can possibly save us troubles with wrapping things that can be used easily without wrapping. 

Wile IOS and Android have sense to be seamless in terms of API because some people can do both platforms cross platform development do not have that restrictions as developers will just most likely work with cordova or react only.

wtrocki:
Closing this PR as follow up work will be done as part of https://github.com/aerogear/aerogear-js-sdk/pull/16 which has more changes.

PR #16, opened 2018-02-25
Multi module platform support.

wtrocki:
For description/history see: https://github.com/aerogear/aerogear-js-sdk/pull/15 
Based on proposal: https://github.com/aerogear/proposals/pull/25

Work in progress, but posting for early visibility.

## Investigation

Following aproach presents number of challenges
- TypeScript exports do not work properly because we expose types instead of typescript code
- Loads of boilerplate in situations when very little changes to platform is required.


packages/metrics-cordova/src/AppMetrics.ts
```diff
@@ -0,0 +1,23 @@
...
+declare var device: any;
+
+/**
+ * Collect application metrics:
+ *
+ *  - appId - application bundle id
+ *  - appVersion - version of the application
+ *  - sdkVersion - AeroGear Services SDK version
+ */
+export class AppMetrics implements Metrics {
```
wei-lee: is there any reason this is cordova specific? The only cordova specific thing to me is the platform. Other values can be used cross-platform?

wtrocki: App version will be as well.

wtrocki: Just experimenting with different solutions so code may be little bit raw in some cases.

wei-lee:
@wtrocki quick question, are we building different bundles for each platform, or are we planning to use the same bundle for all the platforms?

wtrocki:
@wei-lee Great timing for this question. It all depends on distribution we going to settle. We have couple PR's where we need to answer those questions.
We can follow up firebase aproach and have single bundle with `if(cordova) ..` or build separate bundles. Couple investigations happening. 

If you have some opinions about this, this will be the best PR to leave comment:
https://github.com/aerogear/aerogear-js-sdk/pull/14

wtrocki:
Note: Based on suggestions from other PR and list we decided that we do not need additional packages for the cordova/react native. However I would still use this PR for investigations and adapt that to suggestions.

wtrocki:
Closing as cleaned up work available in: https://github.com/aerogear/aerogear-js-sdk/pull/19

PR #18, opened 2018-03-05
Add GUI for Cordova

wtrocki:
## Motivation

Apache Cordova abstract from UI implementations.
Developers can use any framework that has supports mobile device sizes.
We need some UI framework for our demos (change cordova demo to look similar to react native demo). This framework should be simple enough and do not enforce specific development style or tooling as this will make this implementation framework specific. Alternative will be to use ionic, but it seem too heavy/custom for this particular use case.

For this use case I picked JQuery mobile which is:
- Really simple, UI focused framework
- Abstracts from JavaScript implementation (only HTML and CSS)
- Allow developers to reuse code in other framework as it will be generic.

Considering that example application is used only for testing purposes this seem to be all we need. Jquery and JQuery mobile needs to be added on the global scope without browserify. 

## Screenshot

screen shot 2018-03-05 at 7 07 18 pm

## Progress
- [x] Example layout
- [x] Documentation
- [x] Config parsing


david-martin:
I was ready to +1 this, but @wei-lee comment has made me stop & think.
I know @evanshortiss has shared some stuff about ionic recently (ionic deploy),
as has Joe O Reilly (does he have a github account?)
There's also the keyonic example app from @TommyJ1994 https://github.com/TommyJ1994/keyonic-v2.

I think Ionic might be a better choice if we're to choose a ui framework (and js framework in angular) in terms of popularity and familiarity.

wtrocki:
Great. Going to move that PR to ionic

wtrocki:
@david-martin @wei-lee 
Generated ionic application has no cordova structure. I have tried to test that with plain cordova plugins but they do not work properly.
It it's heavily customized and not longer using actual cordova plugins. After using ionic our example will no longer be vanilla Cordova application - it will be ionic application with heavily customized cordova plugins that are imported from `@ionic-native` organization on npm.

Could we bring JQuery mobile for this sprint and post question to the list if we should have Ionic app or plain Cordova app? For plain Cordova app we could use anything that will give us UI like http://www.material-ui.com or even bootstrap css. I have picked JQuery mobile for the moment as it's really stable and hassle free solution.

darahayes:
The nice thing about jQuery mobile is the initial simplicity. It's easy to show off the raw usage of the SDKs with minimal scaffolding. That being said, I think Ionic probably adds a lot more value because it shows 'real world examples' in the context of an Ionic app (i.e. one of the most popular cordova based frameworks). I don't think the same can be said about jQuery mobile anymore.

Additionally, I think an Ionic example would scale better (within the team) when it comes to adding new usage examples of the SDKs over time. (as the SDKs mature)

My concern is that if we start off with a very simple jQuery mobile app, the code becomes hard to read and maintain over time. I think we've all seen this happen with jQuery.

TommyJ1994:
It's still possible to use standard cordova plugins with Ionic v3 as I have done some work recently with both of these, but very much agree with @wtrocki in that its not as user friendly anymore as in Ionic v1. The Ionic native plugins are much better and easier to use and this is the way Ionic are going from now on it seems.

We could use Ionic v1/Angular 1, but Ionic 3/Angular 4/Typescript would have been nicer if the plugin problem wasn't an issue.

TommyJ1994:
Another option that I played with years ago was Onsen UI - https://onsen.io Its pretty similar to Ionic

wtrocki:
Onsen UI will be much better choice as it is just UI framework.

wei-lee:
Onsen UI looks pretty good. Especially you can use either Angular or React or Vue.js which is very nice. Agreed that for this example app we go with a vanilla Cordova app. Onsen UI is a good fit as it is just a UI framework.

Given how much Ionic is now different from a standard Cordova app, we probably will need an example app for it anyway. I think we can use the security template app for it.

wtrocki:
Like suggestions. As we all agreed that Jquery mobile is not the best option - I'm going to close this PR and get some agreement on the list on what we should implement.

PR #20, opened 2018-03-06
Fix Unit tests by adding ts-node

wtrocki:
Fix Unit tests by adding ts-node as dev dependency.
ts-node is required to compile typescript unit tests.

### DIFF
--- a/example/cordova/.gitignore
+++ b/example/cordova/.gitignore
@@ -6,3 +6,4 @@ platforms/*/
 hooks
 merges
 plugins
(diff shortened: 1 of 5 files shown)


wtrocki:
Tests still failing. Checking that now.

wtrocki:
@pb82 After moving to commonjs react native example still works

screen shot 2018-03-06 at 1 17 40 pm

wtrocki:
@pb82 Tested for browserify as well.

PR #23, opened 2018-03-12
fix: path to main module

psturc:
## Motivation
Getting errors in bundler after running `react-native run-android`:
https://gist.github.com/psturc/4aab28264d8ef5e7c7b940d4b015e0a6
## Description
As described in the error, the path specified in `packages/core/package.json` was incorrect.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -2,7 +2,7 @@
   "name": "@aerogearservices/core",
   "version": "1.0.0",
   "description": "Core SDK for AeroGear services",
-  "main": "dist/index.js",
+  "main": "dist/src/index.js",
   "types": "types/index.d.ts",
   "author": "Aerogear <aerogear@googlegroups.com>",
   "license": "Apache-2.0",


packages/core/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@aerogearservices/core",
   "version": "1.0.0",
   "description": "Core SDK for AeroGear services",
-  "main": "dist/index.js",
+  "main": "dist/src/index.js",
```
pb82: ping @wtrocki 

Looks like the typescript compiler puts the resulting file into a src folder instead of creating a bundle in `/dist`. Is this intentional?

wtrocki:
It happens when two folders are included in compilation path. 
Fixed in https://github.com/aerogear/aerogear-js-sdk/pull/24

wtrocki:
@psturc - Closing this PR as it's no longer needed. 
Thank you so much for testing and reporting this issue.

PR #25, opened 2018-03-16
feat: UI for Cordova Example

psturc:
## Motivation
https://issues.jboss.org/browse/AGCORDOVA-210
## Description
Provided UI for Cordova Example App, using http://framework7.io/ HTML framework

## Progress
- [X] Example App
- [X] Demo
- [x] Docs

## Additional Notes

**Demo**: https://giphy.com/gifs/aerogear-cordova-1fnYLFZeYa2dAADqzK

As they are proclaiming on their website, framework7 is iOS specific.
As you can see in the demo it's pretty smooth on iOS device, unfortunately the Android app seems to be pretty laggy. I tried that on my real device (Android 5.0), and it's a bit better, but still not comparable to iOS.

There are quite a lot of discussions about the performance on android with various "fixes", but none of those I've tried significantly helped to boost it https://www.google.com/search?q=framework7+android+slow


example/cordova/www/css/icons.css
```diff
@@ -0,0 +1,70 @@
+/* Material Icons Font (for MD theme) */
```
wtrocki: Let's put that to folder that will allow us to update it etc. For the moment it's not clear where this CSS comes from. Angular Material? It should be clear how to update that.
We could also try to use npm install to distribute that etc.

example/cordova/www/index.html
```diff
@@ -1,50 +1,111 @@
...
-        <script type="text/javascript" src="js/bundle.js"></script>
-    </body>
-</html>
+      </div>
+    </div>
+
+  </div>
+  <!-- Cordova -->
+  <!--
+  <script src="cordova.js"></script>
```
wtrocki: Why this is commented? Just curious if we are aware that this is no longer used.

wtrocki:
@psturc Looks good but it needs Cordova (most important deviceReady event) integration

psturc:
@wtrocki thanks for review, the main issues you mentioned are fixed now.

wtrocki:
Great work! I see that you are still in progress with docs, but I'm really happy to get that merged once you are finished.

psturc:
Docs added, ready to be merged. Wdyt @pb82

pb82:
@psturc testing it now

PR #27, opened 2018-03-27
POC: JS SDK in single package for React Native (and Cordova)

josemigallas:
## Description
It is not possible to have a single, multiplatform JS SDK since it is necessary to access to some native functionality and React Native and Cordova need different and specific implementation for that – React Native needs a [native module](https://facebook.github.io/react-native/docs/native-modules-ios.html) whereas Cordova uses [plugins](https://cordova.apache.org/docs/en/latest/guide/hybrid/plugins/).

The idea so far was to have a common library `aerogear-js-sdk` that would be imported by the user alongside the platform specific one, if needed.

This POC presents an upside-down approach: user imports a single platform-specific library, `aerogear-react-native-sdk` or `areogear-cordova-sdk`, the same exact way that it would work for the Android or iOS SDK. These platform-specific libraries contain all the native bits and they depend on a third and common library `aerogear-js-sdk` that contains JS-only common business logic.

In this POC, the Metrics service has been implemented. The React Native module contains the logic that grabs device and app information and internally uses the JS SDK core module, that sends this data  to the backend app-metrics instance.

## Points of discussion
- What's `platform`? Is it `"android"` or `"android-react"` for instance?
- What's `sdkVersion`? Is it `js-sdk` version or the platform-specific libraries `react-native-sdk`'s?
- Is the base library going to be private? In other words, will `js-sdk` package be able to be downloaded and used standalone by any JS app? Like a web app.
- How are we going to organize the repos? Are we going to have different repos for aerogear-js, react-native-sdk, aerogear-cordova-sdk...?

## TODO
- [x] JS Base SDK (metrics)
- [x] React Native SDK POC (metrics)
- [x] React Native example app implementation
- [ ] React Native SDK in TS
- [ ] React Native SDK test suite
- [x] Cordova SDK POC (metrics)
- [ ] Cordova example app implementation
- [ ] Cordova SDK in TS
- [ ] Cordova SDK test suite

## Usage / Verification
Note that any package has been published yet so dependencies has to be resolved locally and some amount of manual work is required.

1. Since JS SDK's is written in TS, it has to be transpiled in order to RN module to consume it.

```sh
cd aerogear-js-sdk/packages/core
npm install && npm run build
```

2. Clone the RN SDK: https://github.com/josemigallas/aerogear-react-native-sdk at the same level as `aerogear-js-sdk` repo is cloned. Then install dependencies:
```sh
npm install
```

4. Deploy an instance of `metrics-apb` in Openshift and copy its app-metrics URL (more info at https://github.com/aerogear/mobile-core)

5. Add a `mobile-services.json` file in the root folder (same level as package.json) with the correct metrics service configuration. Only `url` and `type` are actually relevant at this point. Example:
```json
{
  "version": 1,
  "clusterName": "",
  "namespace": "",
  "clientId": "",
  "services": [
    {
      "id": "metrics",
      "name": "metrics",
      "type": "metrics",
      "url": "http://aerogear-app-metrics-test.192.168.37.1.nip.io/metrics",
      "config": {}
    }
  ]
}
```

Note that before running on Android, an emulator must be running or a device connected. When running iOS, simulator app should start up automatically.


david-martin:
> What's platform? Is it "android" or "android-react" for instance?
I think the platform would still be android, for metrics purposes

> What's sdkVersion? Is it js-sdk version or the platform-specific libraries react-native-sdk's?
I think sdk version needs to include react-native or cordova

> Is the base library going to be private? In other words, will js-sdk package be able to be downloaded and used standalone by any JS app? Like a web app.
Initially, private makes sense.
If we decide the js sdk can be used for web apps in future, then public later.
Alternatively, we may go with something separate like 'aerogear-web-sdk' that has browser specific code in it, but depends on 'aerogear-js-sdk'

> How are we going to organize the repos? Are we going to have different repos for aerogear-js, react-native-sdk, aerogear-cordova-sdk...?
If we're not going to release aerogear-js-sdk as a separate thing, then all in the 1 repo could make for easier maintenance.

david-martin:
@josemigallas I'd suggest prioritising the Cordova related TODO items given that React Native is not currently on any backlogs/trello boards

pb82:
@josemigallas i agree with Daivd on your first three questions:

> What's platform? Is it "android" or "android-react" for instance?

It should be whatever the platform the user is running on, either Android or iOS. It's easy to get that info in react-native: `import {Platform} from 'react-native'; Platform.OS === 'ios' ...`. In Cordova it seems to be possible with `device.platform`.

> What's sdkVersion? Is it js-sdk version or the platform-specific libraries react-native-sdk's?

Should be the version of the package the user imported, in that case the react-native-sdk. Maybe at some point we want to introduce a policy where we keep the version in sync across all platforms?


> How are we going to organize the repos? Are we going to have different repos for aerogear-js, react-native-sdk, aerogear-cordova-sdk...?

One repo would be ok for now. The react native and cordova plugins can easily import the js sdk by using file:... urls in package.json.

example/react-native/.eslintrc
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "rallycoding",
```
pb82: Is this still a leftover from a different example?

example/react-native/src/components/NavigationController.js
```diff
@@ -0,0 +1,14 @@
+import { StackNavigator, DrawerNavigator } from "react-navigation";
+import { HomeScreen, HttpScreen } from "./screens";
+
+/**
+ * Main navigation controller contains all children screen components
+ */
+const NavigationController = DrawerNavigator({
+  Home: StackNavigator({ screen: HomeScreen }),
```
pb82: Why a `StackNavigator` inside the `DrawerNavigator`. Need the individual screen another navigator?

example/react-native/src/components/screens/HomeScreen.js
```diff
@@ -0,0 +1,70 @@
+import React, { Component } from "react";
+import { View, Text, StatusBar } from "react-native";
+import { RNMetricsService } from "aerogear-react-native-sdk";
```
pb82: Can we call it just `MetricsService`?

example/react-native/src/components/screens/HomeScreen.js
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  state = {
+    res: "",
+    err: ""
+  };
+
+  componentWillMount() {
+    const configuration = {
+      url: "http://aerogear-app-metrics-test.192.168.37.1.nip.io/metrics"
```
pb82: This should ideally come from `mobile-services.json`

example/react-native/src/components/screens/HttpScreen.js
```diff
@@ -0,0 +1,57 @@
+import React, { Component } from "react";
+import { View, Text, StatusBar } from "react-native";
+import { NavDrawerButton } from "../common";
+import { Colors } from "../../assets";
+// import { ConfigService } from "aerogear-core";
```
pb82: any reason to keep those?

example/react-native/src/components/screens/HttpScreen.js
```diff
@@ -0,0 +1,57 @@
...
+
+    return (
+      <View style={styles.containerStyle}>
+        <StatusBar barStyle="light-content" backgroundColor={Colors.primaryDark} />
+        <View>
+          <Text style={styles.welcomeStyle}>
+            HTTP
+          </Text>
+          <Text style={styles.smallStyle}>
+            {/* keycloakConfig = {JSON.stringify(keycloakConfig)} */}
```
pb82: ^^^

packages/core/package.json
```diff
@@ -1,10 +1,15 @@
 {
-  "name": "@aerogearservices/core",
+  "name": "aerogear-core",
```
pb82: Is that ok for you @paolobueno ? I thought we want to use scoped packages?

wtrocki: Will be best to leave scoped packages - this helps a lot with versioning and management.

packages/core/src/metrics/MetricsService.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(configuration: ServiceConfiguration) {
+    this.url = configuration.url;
+  }
+
+  protected sendAppAndDeviceMetrics(metrics: Metrics): Promise<any> {
+    return axios.post(this.url, metrics);
+  }
+
+  protected getClientId(): string {
+    return "453de743207a0232a339a23e5d64b289";
```
pb82: Hradcoded? I think that should refer to https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/utils/ClientIdGenerator.java#L31

paolobueno:
I agree that the sdkVersion should be the version of the js-sdk (maybe we could build a complex string for it if relevant. i.e. `{sdkVersion: "0.1.0, cordova: x.x.x"}` or `{sdkVersion: "0.1.0-cordova"}`).

I also agree that the platformVersion should be the OS version, but if we sync up the versions of all sdks or rather not have a way to distinguish the hybrid js ones we wouldn't be able to tell which one is native and which one is JS.

example/react-native/.eslintrc
paolobueno: Btw I think moving on to this new sdk it might be a nice opportunity to settle on one of the new opinionated tools that avoid jellyfishing, like using semistandard+prettier.

packages/core/types/main.d.ts
```diff
@@ -0,0 +1,52 @@
...
+  public readonly configuration: ServiceConfiguration;
+  constructor();
+  protected sendMetrics(metrics: Metrics): Promise<any>;
+  protected getClientId(): string;
+}
+
+export interface ServiceModule {
+  readonly type: string;
+  readonly configuration: ServiceConfiguration;
+}
```
paolobueno: Isn't this file generated? could be .gitignored

packages/core/package.json
paolobueno: +1 for scoped, we haven't setup the scope in the main npm registry yet, but it is also a good way to distinguish these packages from the existing UPS sdks.

example/react-native/.eslintrc
josemigallas: No @pb82 I just liked this one and use it everywhere now

example/react-native/src/components/NavigationController.js
josemigallas: Yes, this puts the Drawer o top of the Toolbar

example/react-native/src/components/screens/HomeScreen.js
josemigallas: Yes, at the moment this is only to differentiate between the typescript core and the react-native implementation

packages/core/package.json
josemigallas: There wasn't any particular reason, I will change it back.

josemigallas:
@pb82 @david-martin I'll add the files for react-native-sdk too, to rise visibility.

packages/core/types/main.d.ts
josemigallas: I'm not sure about generating the types, that way it can't be used among projects. I have usually always typed them myself, I guess times are changing :P

packages/core-rn/src/metrics/RNMetricsService.js
```diff
@@ -0,0 +1,28 @@
...
+
+const RNMobileCore = NativeModules.MobileCore;
+
+class RNMetricsService extends MetricsService {
+
+  async sendAppAndDeviceMetrics() {
+    const deviceMetrics = await RNMobileCore.getDeviceMetrics();
+    const appMetrics = await RNMobileCore.getAppMetrics();
+
+    appMetrics.sdkVersion = packageJson.version + "-rn";
```
josemigallas: @paolobueno @pb82 @david-martin 
How do you like this approach? It would be sth like `1.0.0-rn` `1.0.0-cordova`.

david-martin: Would it make sense the other way around? e.g. `react-native-1.0.0` and `cordova-1.0.0`

Also, are there changes to the android & ios sdk to conform to this too? e.g. `ios-1.0.0' ?

josemigallas: Good point, right now it's omitted in iOS and Android. On the other hand I wonder if this is really necessary though, since the developer knows already what SDK he's using.

darahayes: > On the other hand I wonder if this is really necessary though, since the developer knows already what SDK he's using.

Big +1 on this point. Personally I would find something like `cordova-1.0.0` or `react-native-1.0.0` unintuitive and confusing, especially when it starts to surface in the dashboards. Also, if users want to build tooling/automation/decision making on top of the raw metrics data, then this format will cause some headaches.

I see two ways of understanding which SDK an app is using:

1. We could assume the developers/owners of the app already know. Perhaps this isn't the best assumption to make.
2. We explicitly add a new field to the App payload that identifies the SDK being used.

wtrocki:
Going to close this PR to not cause confusion.
This work was reused in other PR's (and some react native parts will be filed separately. PR has to wide scope.

PR #29, opened 2018-04-03
Adds base metrics SDK for JS libraries

josemigallas:
## Motivation
https://issues.jboss.org/browse/AGCORDOVA-198

## Description
This is a very minimum value implementation of what will conform the JS base SDK. This base SDK can't be used by itself but presents the common logic that platform-specific SDK need to use (Cordova, RN, web).


wtrocki:
Outdated. See #30

PR #33, opened 2018-04-05
Implement getClientId method in core.

josemigallas:
## Motivation
ID generation should be within common business logic, however the it is stored in the device can differ from platforms.

## Description
`getClientId` is no longer abstract, however it calls internally `getStoredClientId` and `storeClientId` that have to be implemented by dependant SDKs.

## Progress
- [x] Check if there's a stored ID
- [x] Implement clientID generation

## Additional Notes
App examples out of scope at this point since they don't use core.

Test to be added in next PR, alongside other missing tests


packages/core/src/metrics/MetricsService.ts
```diff
@@ -47,4 +42,20 @@ export abstract class MetricsService {
...
     }
 
+    /**
+     * Generates or gets mobile client id
+     */
+    public getClientId(): string {
+        let clientId = this.getStoredClientId();
+
+        if (!clientId) {
+            clientId = "TODO: generate a uuid";
```
josemigallas: @wtrocki would we use here a `UUIDGenerator` class like in https://github.com/feedhenry/fh-sync-js/blob/master/src/clientIdProvider.js#L17 ?

wtrocki: Why not use it then ?

packages/core/src/metrics/MetricsService.ts
```diff
@@ -47,4 +42,20 @@ export abstract class MetricsService {
...
+            this.storeClientId(clientId);
+        }
+
+        return clientId;
+    }
+
+    protected abstract getStoredClientId(): string;
```
wtrocki: My vote will be `getClientId` and `saveClientId`. It could be stored or retriever from somewhere so generic names are better. Also add documentation.

josemigallas: I would still leave 3 methods, 1 that generates a new ID and 2 abstract that get/save. Otherwise the contract between core and core-cordova is broken. How do we ensure it checks if clientId exists before creating a new one? We should leave loading and saving implementation for the specific skds, Cordova will use `window.localStorage` but React Native has to use native persistent storage ([AsyncStorage](https://facebook.github.io/react-native/docs/asyncstorage.html#docsNav)).

wtrocki: I meant to rename them only.

PR #37, opened 2018-04-10
Getting started guide

paolobueno:
## Motivation

Add getting started guide for the JS SDK.

https://issues.jboss.org/browse/AGCORDOVA-200

## Progress
- [x] Getting Started Guide
- [x] Core module skeleton
- [x] Instructions to add mobile-services.json
- [ ] Core module API for Logging?
- [ ] Core module API for Networking?
- [ ] Core module API for Metrics


paolobueno:
@josemigallas @wtrocki I added some stubs for a networking and logging APIs from core, but I'm not sure if they're going to be there. I imagine at least logging should be, not sure if we'll do networking for the JS sdk.

paolobueno:
ping @ciaranRoche

wtrocki:
>  I imagine at least logging should be, not sure if we'll do networking for the JS sdk.

I have no strong opinions about it. Considering how many options they are and what other libraries we going to reuse I think it will be hard to do networking. 
If we use keycloak adapter library for example it comes with their own networking and logging.

docs/core.adoc
```diff
@@ -0,0 +1,31 @@
...
+* Parsing of the configuration file
+* Logging
+* Metrics gathering
+
+== Installation
+
+Use the Cordova CLI to add the Core module as a plugin to your project:
+
+----
+cordova plugin add @aerogearservices/Core
```
josemigallas: Lowercase?

docs/core.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+The Core SDK provides helpers for parsing the configuration file and share common utilities between other SDK modules.
+
+It's main features are:
+
+* Parsing of the configuration file
+* Logging
+* Metrics gathering
+
+== Installation
```
josemigallas: I wouldn't assume for starters that the SDK is only for Cordova. Actually core itself isn't a plugin, maybe we should say that there is a plugin implementation and point to the package/core-cordova readme? For now there will be 2 packages in NPM: `@aerogearservices/core` and `@aerogearservices/core-cordova`.

docs/core.adoc
```diff
@@ -0,0 +1,31 @@
...
+cordova plugin add @aerogearservices/Core
+----
+
+This will install the package from the NPM registry and properly link all native dependencies to the project's cordova Platforms.
+
+Refer to the link:./getting-started[Getting Started] guide for more information on installing the AeroGear SDK.
+
+== Usage
+
+After installation, the Core module will become available through the global `window.MobileCore` object in your project.
```
josemigallas: So finally what should we use as global? I vote for `aerogear`.
`window.aerogear.metrics.MetricsService`
`window.aerogear.sync.login()`

docs/getting-started.adoc
```diff
@@ -1,4 +1,94 @@
-== Getting started
+== Getting Started with AeroGear Services JavaScript SDK
 
-SDK is under construction. 
-No SDK available at the moment.
+This guide provides general information for creating applications that utilize the AeroGear Services JavaScript SDK.
+
+== Prerequisites for Apache Cordova
+
+In order to develop a Cordova application utilizing the SDK you will need the following prerequisites installed:
```
josemigallas: I would assume the user has already a Cordova app, instead of trying to explain the whole process. We offer a service for Cordova apps so I would say `In order to include/use this SDK in your Apache Cordova application...` and omit the `Creating an Apache Cordova app` that doesn't say much.

docs/getting-started.adoc
```diff
@@ -1,4 +1,94 @@
...
+
+ * The link:https://nodejs.org/en/download/[Node.js] runtime and NPM
+ * https://github.com/aerogear/mobile-cli[AeroGear Mobile CLI]
+ * Apache link:https://cordova.apache.org/docs/en/latest/guide/cli/#installing-the-cordova-cli[Cordova CLI]
+
+== Creating an Apache Cordova application
+
+After installing the prerequisites, refer to the link:https://cordova.apache.org/docs/en/latest/guide/cli/index.html#create-the-app[App Creation] guide from Apache Cordova to create your first application.
+
+== Installing the AeroGear JavaScript SDK
```
josemigallas: Since this is a getting-started, what do you think about avoiding as much verbosity as possible? This paragraph doesn't add much, IMO. We could leave the next one with the relevant lines of code, that is `cordova plugin add @aerogearservices/core-cordova`

docs/getting-started.adoc
```diff
@@ -1,4 +1,94 @@
...
+----
+{
+  // ...
+  "dependencies": {
+    "@aerogearservices/core-cordova": "1.0.0",
+    // ...
+  },
+  "cordova": {
+    "plugins": {
+      "cordova-plugin-whitelist": {},
```
josemigallas: Is this plugin here purposely?

docs/getting-started.adoc
```diff
@@ -1,4 +1,94 @@
...
+    <plugin name="core-cordova" spec="1.0.0" />
+</widget>
+----
+
+
+See the link:https://cordova.apache.org/docs/en/latest/platform_plugin_versioning_ref/#plugin-versioning[Apache Cordova documentation for Plugins] for more information.
+
+== Adding the AeroGear Services Configuration file
+
+Utilize the link:https://github.com/aerogear/mobile-cli[AeroGear Mobile CLI] to generate a `mobile-services.json` file and place it inside your project's resource directory.
```
josemigallas: Isn't the mobile-services.json generated (or going to be generated) in the UI in Openshift?

josemigallas:
I don't know how useful would be a networking module and for the logger, I think it would just be a wrapper around console.log. I might be wrong on this though.

docs/core.adoc
wtrocki: `core-cordova` is not valid js identifier. While npm package needs to be named `core-cordova` namespace do not need to as we know that we are in the cordova ecosystem anyway.

ciaranRoche: Awesome, thanks for that @wtrocki makes sense!

josemigallas: @wtrocki But how do we install the plugin from NPM using `@aerogearservices/core` it that's already a package. If we don't want dashes I would go for `corecordova`

docs/core.adoc
paolobueno: This is one thing I'm still not 100% clear on, @josemigallas , should the user install both `core` and `core-cordova` in her project? Or is core-cordova going to use core as a dependency to share the non-native js code?

docs/core.adoc
paolobueno: Yes, with what we're looking at from your work, I'll be changing this to `window.aerogear`

docs/getting-started.adoc
paolobueno: :+1:  I'll go ahead and remove the 'creating an app' section, but for consistency with what's in the android sdk (which lists downloading android studio, etc as prerequisites), I was thinking of actually keeping this part.

docs/core.adoc
josemigallas: Yes that's the goal here, to use only and exclusively `core-cordova`.

paolobueno:
@wtrocki sure, thanks I'll leave the empty headings for other modules then, even though they might not exist

PR #40, opened 2018-04-13
Adds missing metric in Android module

josemigallas:
## Description
`sdkVersion` metrics must be added by the JavaScript interface, however for explicitness it should be added in the Android module. This way if for some reason the value is not added by JS code, the element will still be there in the metrics payload.

The way this metrics is set is something like:
```javascript
function sendAppAndDeviceMetrics() {
  this.getAppAndDeviceMetricsFromNativeModule()
    .then(metrics => {
      metrics.app.sdkVersion = this.getPluginVersionFromWherever();

      // publish
    });
}

### DIFF
--- a/packages/core-cordova/src/android/MobileCoreModule.java
+++ b/packages/core-cordova/src/android/MobileCoreModule.java
@@ -71,6 +71,8 @@ public class MobileCoreModule extends CordovaPlugin {
 
     final JSONObject appMetrics = new JSONObject();
     appMetrics.put("appId", packageName);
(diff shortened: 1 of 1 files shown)


wtrocki:
I see that change is not complete. I guess we can merge it, but I think that we can easily provide complete implementation as you have mentioned. I'm worried if we start PR with some unfinished changes we may then forget about this or leave some things without implementation.

josemigallas:
@wtrocki It is neither incomplete nor failing, this change is optional and doesn't have anything to do with JS implementation. It doesn't change the behaviour of the SDK.

PR #41, opened 2018-04-13
Use Core package in Cordova plugin

josemigallas:
## Motivation
We need to have the common business logic from the actual platform implementation. This way, the JS-Core SDK will be extensible to any other JS-based platforms.

## Description
Making `core` a Cordova plugin itself allows it to be imported from `core-cordova` and its classes to be used and extended.

Core will be both a Cordova plugin and a NPM package. For the plugin to work, the transpiled files will also be bundled into a `www/bundle.js` that will be required by means of cordova from `core-cordova` once the dependency is installed:

```javascript
// CordovaMetricsSerice.js in core-cordova
const MetricsService = cordova.require("core.core").MetricsService;

class CordovaMetricsServices extends MetricsService {

}

// It is exported to window as MetricsService, so that it makes the cordova implementation transparent
exports.MetricsServices = CordovaMetricsService;
```
In a Cordova application, the SDK would be used like this:
```javascript
<script>
    var config = { url: "http://192.168.1.6:3000/metrics" };
    var metricsService = new aerogear.MetricsService(config);      
    var metricsService.sendAppAndDeviceMetrics()
        .then( ... );
</script>
```
## TODO
- [x] use core in cordova-core

## Next

Although this PR presents a workable solution, there is still some work to do in order to reach a final state:
- Testing
        Since `core` is required during runtime by cordova.require, tests should be included in the example application. Functionality like getting/saving the clientId could be tested.
- Documentation
        There's a big lack of documentation at this point. @paolobueno is committed to it if I'm not mistaken
- Plugin architecture
        At the moment the whole plugin fits into a single class, due its simplicity. This should be improved though, establishing a proper architecture (and index entry point, a folder per module, etc)
- TS/typings
        For the sake of quality, maintainability and extensibility it would be good to have the code base written in TS and then transpiled into `www/`.
- Package names
        `core` and `core-cordova` plugin names/ids needs to be determined.
- General availabilty
        At the moment the package has to be used locally, once we come up with a proper name, they can be published and the local references removed.


wtrocki:
> Package names

We could drop some email to the list to get that established over the time.

example/cordova/www/pages/metrics.html
```diff
@@ -45,20 +45,31 @@ <h4>Response:</h4>
   return {
     beforeCreate() {
       var self = this;
-      self.$("#metrics-info").html("No metrics");
-      window.MobileCore.MetricsService.getMetrics(
-        function (metrics) {
+      self.$("#metrics-info").html("Fetching metrics...");
+
+      self.metricsService = new aerogear.MetricsService({ url: "http://192.168.1.6:3000/metrics" });
```
wtrocki: However this example works it is not what other SDK is doing. Action is that Metrics are instantiated and are sending data to server on start.

wtrocki: We are not using config here. We should use config parser.

example/cordova/www/pages/metrics.html
```diff
@@ -45,20 +45,31 @@ <h4>Response:</h4>
...
+      self.metricsService.getAppAndDeviceMetrics()
```
wtrocki: Method should be `sendAppAndDeviceMetrics`

packages/core-cordova/www/metrics.js
```diff
@@ -0,0 +1,57 @@
...
+    this.id = id;
+    this.data = data;
+  }
+
+  collect() {
+    return this.data;
+  }
+}
+
+class CordovaMetricsService extends MetricsService {
```
wtrocki: [Optional] Let's add documentation explaining what is happening here.

packages/core-cordova/www/metrics.js
```diff
@@ -0,0 +1,57 @@
...
+
+  getAppAndDeviceMetrics() {
+    return new Promise((res, rej) => {
+      cordova.exec(
+        metrics => {
+          const plugins = cordova.require("cordova/plugin_list");
+          const sdkVersion = plugins.metadata["core-cordova"];
+          metrics.app.sdkVersion = sdkVersion;
```
wtrocki: SDK version should be done inside the app metrics.

packages/core-cordova/www/metrics.js
```diff
@@ -0,0 +1,57 @@
...
+        const appMetrics = new MetricsImpl("app", metrics.app);
+        const deviceMetrics = new MetricsImpl("device", metrics.device);
+
+        return this.publish([appMetrics, deviceMetrics]);
+      });
+  }
+
+  getSavedClientId() {
+    // TODO: get from device storage or return undefined
+    return undefined;
```
wtrocki: Do we have ticket to do it? if not let's implement that here.

example/cordova/www/pages/metrics.html
josemigallas: True, I'll add it in the todo list

packages/core-cordova/www/metrics.js
josemigallas: app metrics is just a plain JSON returned by the Android module, are you suggesting we add a wrapper class around it? IMO we could leave it like this for now, to keep things a little simpler.

wtrocki: Yeah. I have noticed that we wrap that payload bellow into interfaces.

wtrocki:
@josemigallas Tried that locally. I will wait for the rest of the changes and give it a final spin.

packages/core-cordova/www/metrics.js
paolobueno: Don't you think we can start branching some of these todo items to tickets? I think getting the main structure merged should be a priority.

.gitignore
```diff
@@ -73,3 +73,4 @@ package-lock.json
 ## Dist folders
 dist
 types
+**/bundle.js
```
paolobueno: Maybe add all `www/`s if we end up having more than 1 file there in the future.

example/cordova/www/pages/metrics.html
```diff
@@ -45,20 +45,31 @@ <h4>Response:</h4>
...
+        .then(metrics => {
           self.$("#metrics-info").html(JSON.stringify(metrics));
```
paolobueno: Haven't checked how it looks currently, but the json should be wrapped in `<pre><code>`

packages/core-cordova/www/metrics.js
```diff
@@ -0,0 +1,57 @@
+const MetricsService = cordova.require("core.core").MetricsService;
```
paolobueno: Is this going to stay as "core.core" ? We might need to revise this when trying to publish the plugin.

packages/core-cordova/www/metrics.js
paolobueno: If we look at the roadmap, getting the app metrics working is in a [future card](https://trello.com/c/p2gBN9XB/48-5-cordova-appmetrics-as-a-mobile-developer-i-want-the-cordova-sdk-to-give-me-app-launch-metrics-such-that-i-can-visualise-and-an), not saying we can't speed things up, of course :)

wtrocki: @paolobueno Good point! Let's create tickets to make sure that will be covered.

packages/core-cordova/www/metrics.js
wtrocki: Yeah. We pushed that to the list so it will change.

example/cordova/www/pages/metrics.html
josemigallas: It looks horrible hehe, this actually could or should be removed and let only sendAppAndDeviceMetrics

packages/core-cordova/www/metrics.js
josemigallas: The first core is the plugin, the second is the module. It could be "aerogear.core"

packages/core-cordova/www/metrics.js
josemigallas: Is it too much now maybe?

josemigallas:
@wtrocki @paolobueno I removed the todos and created JIRAs instead. As Paolo pointed out, this is enough to meet our acceptance criteria (alongside with the new tickets) so I really wouldn't worry to much about the example app or other technicalities at this point.

I would however first merge #42 and rebase this one before merging.

wtrocki:
@josemigallas Ready for rebase. Going to give it a final run!

PR #44, opened 2018-04-18
WIP Spike: Single core pkg

paolobueno:
## Motivation
Avoid the need for separate packages for cordova and RN

## Description

- Fuse current core-cordova with core
- Use `browserify --standalone` to change bundle module type
- Change class namespace and add skeleton for RN support (build.gradle, etc)

Added an extra `org.aerogear.mobile.js.core` to avoid being in the same namespace as the native SDK, but I guess that's optional.
Skeleton for RN inspired by https://github.com/frostney/react-native-create-library

## Progress
- [X] Add RN skel
- [x] Refactor cordova and RN-specific wrapper code to different classes
- [x] Figure out structure for shared and specific TS
- [x] Figure out structure for shared and specific JS


paolobueno:
Currently `cordova build/run android` works but I'm getting "Class not found" in the metrics screen, couldn't figure this out yet.

wtrocki:
@paolobueno No worries. We just want to get idea first. We can resolve issues together.

PR #45, opened 2018-04-18
AEROGEAR-2440 added separate auth package

StephenCoady:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-2440

## Description
Added standalone auth package. This enables a mobile application to consume auth as a standalone service.

## Additional Notes
Tested using a new example application created here: https://github.com/StephenCoady/cordova-showcase-template

## Verification
If you wish, this package can be verified by using the above linked application. To do so:

* cd packages/auth
* npm link @aerogearservices/auth
* npm install
* npm run build

Change directory to the showcase application and follow the steps to run in the readme. You should then be able to consume the auth package and login/logout/get current profile as per the ticket.


packages/auth/package.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "@aerogearservices/auth",
```
wei-lee: should this just be `@aerogear` based on the latest discussion on the mailing list.

packages/auth/src/AuthService.ts
```diff
@@ -0,0 +1,44 @@
+import Keycloak from "keycloak-js";
+
+export class AuthService{
+    static auth: any = {};
+
+    loadUserProfile(): any {
```
wei-lee: I think we should create a type for the return `UserProfile` to make it easier for developers to work with, and provide documentations.

packages/auth/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "devDependencies": {
+    "@types/chai": "4.1.2",
+    "@types/chai-as-promised": "7.1.0",
+    "@types/cordova-plugin-camera": "2.4.0",
+    "@types/lodash.find": "^4.6.3",
+    "@types/mocha": "^5.0.0",
+    "@types/node": "^9.4.6",
+    "@types/proxyquire": "1.3.28",
+    "@types/sinon": "^4.3.1",
+    "@types/uuid": "^3.4.3",
```
wtrocki: You do not use this package along with lodash.find types above. Let's remove that.

packages/auth/src/AuthService.ts
wtrocki: https://github.com/keycloak/keycloak-js-bower/blob/master/dist/keycloak.d.ts#L181-L191

wtrocki: Any is bad. If we do wrapper let's not make it worse than official lib:

wtrocki: Any is bad. If we do wrapper let's not make it worse than official lib:
https://github.com/keycloak/keycloak-js-bower/blob/master/dist/keycloak.d.ts#L476

Here is user interface that @wei-lee is mentioning:
https://github.com/keycloak/keycloak-js-bower/blob/master/dist/keycloak.d.ts#L181-L191

packages/auth/src/AuthService.ts
```diff
@@ -0,0 +1,44 @@
...
+        // Retrieve User Profile
+        return new Promise((resolve, reject) => {
+            AuthService.auth.authz.loadUserProfile().success((profile: any) => {
```
wtrocki: This is not going to work. What we probably should od is:
` return AuthService.auth.authz.loadUserProfile()` and remove all bellow

packages/auth/src/AuthService.ts
```diff
@@ -0,0 +1,44 @@
...
+        AuthService.auth.authz.logout();
+    }
+
+    isAuthenticated(): boolean{        
+        return AuthService.auth.authz.authenticated;
+    }
+
+    static init(keycloakConfig: any) : Promise<any> {
+        if(keycloakConfig){
+            let keycloak = Keycloak(keycloakConfig);
```
wtrocki: This looks like copy from js code.

packages/auth/src/AuthService.ts
```diff
@@ -0,0 +1,44 @@
...
+    isAuthenticated(): boolean{        
+        return AuthService.auth.authz.authenticated;
+    }
+
+    static init(keycloakConfig: any) : Promise<any> {
+        if(keycloakConfig){
+            let keycloak = Keycloak(keycloakConfig);            
+            return new Promise((resolve, reject) => {
+                keycloak.init({ onLoad: 'check-sso' }).success(() => {
+                    AuthService.auth.authz = keycloak;
```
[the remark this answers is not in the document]
wtrocki: Preferable is to use self or enclosed private instance.

wtrocki:
This looks how it should look like. I have added couple comments that this should make this more clean in terms of style. Good work!

packages/auth/src/AuthService.ts
```diff
@@ -0,0 +1,44 @@
...
+        AuthService.auth.authz.login();
+    }
+
+    logout(): void{
+        AuthService.auth.authz.logout();
+    }
+
+    isAuthenticated(): boolean{        
+        return AuthService.auth.authz.authenticated;
+    }
```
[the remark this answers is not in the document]
wtrocki: Moving to constructor or extending will resolve this problem.

wtrocki:
@StephenCoady Created separate PR with my suggestion:
https://github.com/StephenCoady/aerogear-js-sdk/pull/1

packages/auth/src/AuthService.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Wrapper class for {Keycloak.KeycloakInstance}
+ */
+export class AuthService {
+
+    private auth: KeycloakInstance;
+
+    constructor(config: KeycloakInitOptions) {
+      this.auth = Keycloak(config);
```
paolobueno: Are you sure this doesn't need `new`? It's possible if the lib is a bit weird.

wtrocki: @paolobueno It is actually needing new, but typescript types do not allow that.

packages/auth/src/AuthService.ts
```diff
@@ -0,0 +1,63 @@
...
+		 * Redirects to logout.
+		 * @param options Logout options.
+		 * @param options.redirectUri Specifies the uri to redirect to after logout.
+		 */
+    public logout(): KeycloakPromise<void, void> {
+      return this.auth.logout();
+    }
+
+    public isAuthenticated(): boolean | undefined {
+        return this.auth.authenticated;
```
paolobueno: Is there a reason for returning `undefined` here? If it's because `auth` might be null then it might be better to `if (!this.auth) {return false;}` beforehand.

Also use `boolean?` for nullable types.

packages/auth/src/AuthService.ts
```diff
@@ -0,0 +1,63 @@
...
+        initOptions.onLoad = "check-sso";
+      }
+      return this.auth.init(initOptions);
+    }
+
+    /**
+		 * Loads the user's profile.
+		 * @returns A promise to set functions to be invoked on success or error.
+		 */
+    public loadUserProfile(): KeycloakPromise<KeycloakProfile, void> {
```
paolobueno: -1 for exposing a custom promise type, is it needed? does it have any extra useful methods? Maybe just return a `Promise` if it's castable.

packages/auth/src/AuthService.ts
```diff
@@ -0,0 +1,63 @@
+import Keycloak from "keycloak-js";
+import {KeycloakError, KeycloakInitOptions, KeycloakInstance, KeycloakProfile, KeycloakPromise} from "keycloak-js";
```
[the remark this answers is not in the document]
wtrocki: Yeah. I prefer to split them :)

packages/auth/src/AuthService.ts
wtrocki: It was just copy paste from keycloak types.

wtrocki:
@StephenCoady Keeping it on your branch will make it hard for me to work with (cannot push to this branch somehow). Going to apply all the changes and migrate this upstream. Is that ok?

wtrocki:
Going to create new PR, with all suggestions.

PR #47, opened 2018-04-18
Remove example apps

wtrocki:
## Motivation

Removing example applications (they will be still available in the previous tag).
This is aligned with decision made on the list to use separate repositories with templates for cordova and react.


wtrocki:
Follow up work with documentation will be done as separate PR.

PR #49, opened 2018-04-19
Move to bluebird

wtrocki:
## Motivation

Use bluebird for the Auth module


packages/auth/src/AuthService.ts
```diff
@@ -17,40 +18,48 @@ export class AuthService {
...
+  public logout(): Bluebird<void> {
+    return new Bluebird((resolve, reject) => {
+      return this.auth.logout().error(reject).success(resolve);
+    });
   }
 
-  public isAuthenticated(): boolean | undefined {
-    return this.auth.authenticated;
+  public isAuthenticated(): boolean  {
+    return !!this.auth.authenticated;
```
josemigallas: Why the double negation?

packages/auth/package.json
```diff
@@ -61,6 +60,7 @@
     "typescript": "^2.8.1"
   },
   "dependencies": {
+    "bluebird": "^3.5.1",
```
josemigallas: Typescript comes already with built-in promises, why rely on a 3rd party module?

packages/auth/src/AuthService.ts
wtrocki: Addresses previous issue. Unwraps undefined.

josemigallas: Could we do `this.auth.authenticated as boolean` instead?

josemigallas: Or even `return value ? true : false;` if the type must be really boolean

packages/auth/package.json
wtrocki: @josemigallas This is designed to work with the web and needs polyfil. Tried typescript promises and they are not polyfilled.

packages/auth/src/AuthService.ts
wtrocki: @josemigallas This comes wrong from keycloak. `return value ? true : false;` has the same effect. Let's not go into holly war with the style on PR.

PR #50, opened 2018-04-20
Separation for native plugin

wtrocki:
WIP

Idea of this aproach is to follow the keycloak and firebase (and current proposal) so javascript code will be totally separated from the native and used directly.
Native functionalities will be available as separate plugins that can be called from js code when platform is detected.


packages/core-cordova/README.md
```diff
@@ -1,16 +1,13 @@
...
 
-TODO
+## Installation
 
-<!-- You can install this plugin with Cordova CLI, from npm:
+You can install this plugin with Cordova CLI, from npm:
 
 ```
-$ cordova plugin add @aerogearservices/core-cordova
+$ cordova plugin add @aerogearservices/cordova-plugin-metrics
```
paolobueno: Do you think we should have a separate plugin for each native feature?

wtrocki: Done review this morning over the top level requirements around platforms.

PR #51, opened 2018-04-20
WIP Split single package spike into multiple  packages

paolobueno:
## Motivation

Follow up from #44 but using a separate core-cordova and core-rn packages.


wtrocki:
@paolobueno Reviewed that locally. Let's have chat over what can be taken.

PR #52, opened 2018-04-20
Separation for native plugin

wtrocki:
## Motivation

Idea of this aproach is to follow the keycloak and firebase (and current proposal) so javascript code will be totally separated from the native and used directly.
Native functionalities will be available as separate plugins that can be called from js code when platform is detected.


wtrocki:
@paolobueno @josemigallas Added changes that made this stuff working. 

I'm going now to try integrate that with the demo app and we could to 0.1.0 test release for both auth and core. The metrics work will be done as part of this epic: https://issues.jboss.org/browse/AGCORDOVA-221 This should be done for the 

What is outstanding here is jst documentation that could be done in separate PR.

packages/core/src/metrics/MetricsService.ts
```diff
@@ -1,72 +1,107 @@
 import uuid from "uuid/v1";
 import { ServiceConfiguration } from "../configuration";
-import { Metrics, MetricsPayload, MetricsType } from "./model";
+import { Metrics, MetricsPayload } from "./model";
+import { CordovaAppMetrics } from "./platform/CordovaAppMetrics";
+import { CordovaDeviceMetrics } from "./platform/CordovaDeviceMetrics";
 import { MetricsPublisher, NetworkMetricsPublisher } from "./publisher";
+import { isMobileCordova } from "./platform/PlatformUtils";
```
paolobueno: :+1:

packages/core/src/metrics/MetricsService.ts
```diff
@@ -1,72 +1,107 @@
...
-
-        return this.publisher.publish(payload);
+  /**
+   * Publish metrics using predefined publisher
+   *
+   * @param type type of the metrics to be published
+   * @param metrics metrics instances that should be published
+   */
+  public publish(type: string, metrics: Metrics[]): Promise<any> {
+    if (!type) {
```
paolobueno: can replace this with an [enum](https://www.typescriptlang.org/docs/handbook/enums.html) in the future

wtrocki: Actually we had that as interface before. I removed that (see left side).
It was because the metrics side doesn't know what other SDK will use as key and defining that in core will mean that people will need to update core to use and pass new type which is really bad.

packages/core/src/metrics/platform/PlatformUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Detect React Native.
+ *
+ * @return {boolean} True if ReactNative environment is detected.
+ */
+export const isReactNative = function() {
+  return (
+    // tslint:disable-next-line:no-string-literal
+    typeof navigator === "object" && navigator["product"] === "ReactNative"
```
paolobueno: great stuff, also for detecting if there's a specific native module on I think we can use
```typescript
import {NativeModules} from 'react-native';
if(NativeModules["ModuleName"]){
...
})
```

wtrocki:
I have tested this implementation with sample application. 
There will be some major changes in end user api needed that we could improve in other PR, but apart from that everything should work. To know what I'm talking about I have made PR's with TODO's so we can actually track those and create jiras/or fix things on the spot. 

@josemigallas Configparser inteliassist is not working with any so we should probably remove the type. It makes no sense and helper methods should be restored.

josemigallas:
@wtrocki if type intellisense does not work, I think it's a typings problem. I suggest we create an enum instead but helper methods would also be fine if you guys prefer that approach.

wtrocki:
@josemigallas See example app integration. Config should be an function not class per se. It delivers no value and it's clunky to use in plain JS.

PR #53, opened 2018-04-23
Revert config helper methods

josemigallas:
## Description
Revert to the helper methods `getMetricsConfig()`, etc. 

`ServiceType` is now an enum that define the service type names as constants.

### DIFF
--- a/packages/core/src/configuration/ConfigurationParser.ts
+++ b/packages/core/src/configuration/ConfigurationParser.ts
@@ -4,7 +4,11 @@ import { AeroGearConfiguration, ServiceConfiguration } from "./";
 /**
  * List of types of all supported services.
  */
-export type ServiceType = "metrics" | "keycloak" | "push" | any;
+export enum ServiceType {
+  METRICS = "metrics",
+  KEYCLOAK = "keycloak",
+  PUSH = "push"
+}
 
 /**
  * Represents a configuration parser.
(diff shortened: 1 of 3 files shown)


packages/core/src/configuration/ConfigurationParser.ts
```diff
@@ -4,7 +4,11 @@ import { AeroGearConfiguration, ServiceConfiguration } from "./";
 /**
  * List of types of all supported services.
  */
-export type ServiceType = "metrics" | "keycloak" | "push" | any;
+export enum ServiceType {
+  METRICS = "metrics",
+  KEYCLOAK = "keycloak",
+  PUSH = "push"
```
wei-lee: I don't think the core sdk need to know all the service types. It means we have to update this file every time when there is a new service is added. It also means when we change the name/type of a service module, the core module needs to be updated as well. Does it really needed?

I think the best place for those value should be in each service module itself.

packages/core/src/configuration/ConfigurationParser.ts
```diff
@@ -21,7 +25,33 @@ export class ConfigurationParser {
...
+   * Get Keycloak service configuration object
+   */
+  public getKeycloakConfig(): ServiceConfiguration | undefined {
+    return this.getConfig(ServiceType.KEYCLOAK);
+  }
+
+  /**
+   * Get Push service configuration object
+   */
+  public getPushConfig(): ServiceConfiguration | undefined {
```
wei-lee: I think those methods provides very little benefit, but cause big problem for maintaining.

wtrocki: What problems you mean? This class has no functional value if we do not provide this methods. It is not doing any parsing. Helper methods are all that we could done here.

packages/core/src/configuration/ConfigurationParser.ts
wtrocki: @wei-lee @josemigallas how about

josemigallas: You're probably right, maintainability-wise this is not the right approach.

packages/core/src/configuration/ConfigurationParser.ts
josemigallas: Its functionality really is to abstract us from the JSON file itself, so that we should make it as implicit as possible, it is a helper class after all.

packages/core/src/configuration/ConfigurationParser.ts
wei-lee: yes, I think this is better.

packages/core/src/configuration/ConfigurationParser.ts
wei-lee: The problem I have is that every time when a new service is added, should we add a new helper class here? If so it means we have to add new helper here (1 line of code), bump version and release a new version and maybe update all other modules to use this new version. It is quite a lot of maintenance work, but with not much benefit. It is enough to have the `getServiceConfig` method.

wtrocki: @wei-lee That's awesome. We are going to update that then before we release this! Thank you so much for suggestion. It will be better for end users.

josemigallas: Fair enough

PR #55, opened 2018-04-23
rename parser and remove helper methods

josemigallas:
## Description
First change, remove helper methods and create a static `ID` property to services.

Since they are already services, I thought plain ID would be less redundant.
```
MetricsService.ID
MetricsService.SERVICE_ID
```

## Next
ConfigHelper will be used exclusively inside services, user won't have to use it.


wtrocki:
@josemigallas I think to fully accept this we need to get this 

> ConfigHelper will be used exclusively inside services, user won't have to use it.

packages/core/src/configuration/ConfigurationHelper.ts
```diff
@@ -0,0 +1,26 @@
...
+  constructor(config: AeroGearConfiguration) {
+    this.configurations = config.services || [];
+  }
+
+  /**
+   * Get a service configuration object, provided an existing id is given
+   * @param id string - The id of the service
+   */
+  public getConfig(id: string): ServiceConfiguration | undefined {
+    return find(this.configurations, service => service.id === id);
```
wei-lee: just to avoid potential problem of using uppercase/lowercase letters, let's ignore the cases here.

josemigallas: I would say the opposite, we should consider that an error IMO. An id should be unique and we should expect our service ids are standardize, that's why we define this `MyService.ID` constant, to avoid case errors. @wtrocki wdty?

wei-lee: My thoughts are:

1. there is nothing stop developers from just passing a string. I think it should be allowed.
2. there could be custom services added in the future, and we may not be able to control the id value there. 

I think we should align this with the current naming convention on the server side. Are we ignoring cases on the backend? @maleck13 any idea?

wtrocki: This will be only inside of our SDK's so if there is case issue it should fail badly and we should correct that.

PR #57, opened 2018-04-23
no config helper from user side

josemigallas:
## Description
Removes the Config helper class from user side

### DIFF
--- a/packages/auth/src/AuthService.ts
+++ b/packages/auth/src/AuthService.ts
@@ -1,3 +1,4 @@
+import { AeroGearConfiguration, ConfigurationHelper, ServiceConfiguration } from "@aerogear/core";
 import Bluebird from "bluebird";
 import Keycloak from "keycloak-js";
 import { KeycloakError, KeycloakInitOptions, KeycloakInstance, KeycloakProfile, KeycloakPromise } from "keycloak-js";
@@ -11,8 +12,18 @@ export class AuthService {
 
   private auth: KeycloakInstance;
 
-  constructor(config: any) {
-    this.auth = Keycloak(config);
+  constructor(appConfig: AeroGearConfiguration) {
+    const configuration = new ConfigurationHelper(appConfig).getConfig(AuthService.ID);
+    let internalConfig;
+
+    if (!configuration) {
+      console.info("Keycloak configuration is missing. Authentication will not work properly.");
+      internalConfig = {};
+    } else {
+      internalConfig = configuration.config;
+    }
+
+    this.auth = Keycloak(internalConfig);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/core/src/metrics/MetricsService.ts
```diff
@@ -19,10 +19,22 @@ export class MetricsService {
 
   private publisher: MetricsPublisher;
 
+  private readonly configuration: ServiceConfiguration;
   private readonly defaultMetrics: Metrics[];
 
-  constructor(private readonly configuration: ServiceConfiguration) {
-    this.publisher = new NetworkMetricsPublisher(configuration.url);
+  constructor(appConfig: AeroGearConfiguration) {
+    const configuration = new ConfigurationHelper(appConfig).getConfig(MetricsService.ID);
```
wtrocki: Maybe we could move that to method. It makes no sense to create object that is used instantly.

packages/core/src/metrics/MetricsService.ts
```diff
@@ -19,10 +19,22 @@ export class MetricsService {
...
+
+    if (!configuration) {
+      console.info("Metrics configuration is missing. Metrics will not work properly.");
```
wtrocki: Replace " Metrics will not work properly" => " Metrics will not be published to remote server"

packages/core/src/metrics/publisher/EmptyMetricsPublisher.ts
```diff
@@ -0,0 +1,12 @@
+import { MetricsPayload } from "../model";
+import { MetricsPublisher } from "./MetricsPublisher";
+
+export class EmptyMetricsPublisher implements MetricsPublisher {
+
+  public publish(metrics: MetricsPayload): Promise<any> {
+    return new Promise((res, rej) => {
+      res();
```
wtrocki: Let's add console.log :)

wtrocki:
PR changing the example app: 
https://github.com/aerogear/cordova-showcase-template/pull/6

packages/core/src/metrics/MetricsService.ts
paolobueno: +1, I think we can move the stuff in `ConfigurationHelper` to the `AeroGearConfiguration` class

packages/auth/src/AuthService.ts
```diff
@@ -11,8 +12,18 @@ export class AuthService {
...
   private auth: KeycloakInstance;
 
-  constructor(config: any) {
-    this.auth = Keycloak(config);
+  constructor(appConfig: AeroGearConfiguration) {
+    const configuration = new ConfigurationHelper(appConfig).getConfig(AuthService.ID);
+    let internalConfig;
+
+    if (!configuration) {
+      console.info("Keycloak configuration is missing. Authentication will not work properly.");
```
paolobueno: `console.error()`

paolobueno: `console.warn()` or `error()`

josemigallas: Agree

PR #39, opened 2018-04-13
Document Auth SDK

TommyJ1994:
## Motivation
Document how to add the Cordova SDK (IDM) to a Cordova App and use it so developers have instructions on how to use the SDK

### Description
Create Documentation for the Auth SDK.

PR #58, opened 2018-04-24
docs: Update documentation with final module shape and usage

paolobueno:
## Motivation
Update core.adoc with current sdk structure


packages/core/README.md
```diff
@@ -1,4 +0,0 @@
-  TODO DOCS
```
paolobueno: I realized this would look very similar to a converted version of core.adoc.

Personally I'd like to avoid duplicating content, but the package's readme is going to be visible in NPM. Should it maybe contain a link to the hosted documentation? Not sure.

README.md
```diff
@@ -15,8 +23,7 @@ AeroGear SDK for Java Script, Apache Cordova and React Native.
 
 AeroGear Services SDK consist of set of separate SDKs
 
-- [Core](./docs/core/README.adoc): Common base for all SDKs
-- [Metrics](./docs/metrics/README.adoc): Mobile application metrics SDK
+- [Core](./docs/core.adoc): Common base for all SDKs
```
josemigallas: @wtrocki @paolobueno are we going to have metrics separated in the end? Let's add `Auth` too, no?

docs/core.adoc
```diff
@@ -10,22 +10,60 @@ It's main features are:
 
 == Installation
 
-Use the Cordova CLI to add the Core module as a plugin to your project:
+The `@aerogear/core` module is available via NPM:
 
 ----
-cordova plugin add @aerogearservices/core-cordova
+npm install @aerogear/core
```
josemigallas: `--save` :P

docs/core.adoc
```diff
@@ -10,22 +10,60 @@ It's main features are:
...
+----
+
+This will install the packages with these native dependencies from the NPM registry and properly link all native dependencies to the project's cordova Platforms.
 
 Refer to the link:./getting-started[Getting Started] guide for more information on installing the AeroGear SDK.
 
 == Usage
 
-After installation, the Core module will become available through the global `window.aerogear` object in your project.
+The Core module supports parsing of the configuration file generated by the mobile CLI:
```
josemigallas: This has changed a little since yesterday, now the parser will be transparent for the user. It will only be required to import the JSON file and pass it to the constructor:
```js
import { MyService } from "@aerogear/my-module";
import appConfig from "../mobile-services.json";

const service = new MyService(appConfig);
```

docs/core.adoc
```diff
@@ -10,22 +10,60 @@ It's main features are:
...
+import configuration from '../mobile-services.json';
+const metrics = new MetricsService(configuration);
+metrics.sendAppAndDeviceMetrics().then(() => {
+  console.info("Metrics registered successfully");
+});
+----
+
+App metrics record:
+
+- *ID*: The identifier of your App
```
josemigallas: I think this one is `clientId`.

packages/core/README.md
josemigallas: Yeah it would be more user friendly to have the info right on NPM, we could simply put some usage instructions and API reference in the Readme, maybe a generated js-doc.

docs/core.adoc
wtrocki: Depending on the version of NPM save will be default option.

README.md
wtrocki: Metrics is separate docs.

packages/core/README.md
wtrocki: Let's leave that for now. Ideally readme should just point to documentation to not replicate content. I have the same issue on IOS. Cocoapods

README.md
ciaranRoche: @wtrocki We have a PR here covering auth documentation https://github.com/aerogear/aerogear-js-sdk/pull/39

wtrocki: Great. Let's split metrics then.

docs/core.adoc
wtrocki: @josemigallas Let's propose some update here to simplify @paolobueno work

docs/core.adoc
wtrocki: Let's not include that. This is internal implementation detail.

docs/core.adoc
josemigallas: That's an even better reason to add it.

paolobueno: `--save` has been on by default since npm 5, and latest nodejs LTS (carbon) uses 5.6: https://nodejs.org/en/download/releases/.

Current `npm help install` doesn't even list `--save` as a valid flag anymore, just the different flavors (`--save-dev, --no-save, etc`) so my vote is to keep it out of new documentation.

docs/core.adoc
paolobueno: :+1: I guess there's really not much 'usage' for the core module, I'll reword this as I split away the metrics part

docs/core.adoc
```diff
@@ -10,22 +10,34 @@ It's main features are:
...
 
-=== Networking
+[source, javascript]
+----
+import { MetricsService } from '@aerogear/core';
+import configuration from '../mobile-services.json';
 
-=== Metrics Service
+// utilizes Core to retrieve it's own configuration
+const metrics = new MetricsService(configuration);
```
paolobueno: @josemigallas @wtrocki  I've tried to change the Usage section to point out how core's used by other modules and doesn't make sense by itself but it can be a bit clunky, removing this section is also an option.

PR #59, opened 2018-04-24
Add has realm functions

ciaranRoche:
## Motivation
Currently, we offer no way for a developer to check the Realm Role or get a list of Roles associated with a user.
## Description
This PR adds both a `hasRealmRole()` and a `getRealmRole()` functions.
## Verify
There is a current PR on the showcase app using the above functions https://github.com/aerogear/cordova-showcase-template/pull/7


packages/auth/src/AuthService.ts
```diff
@@ -68,4 +68,22 @@ export class AuthService {
...
+  public hasRealmRole(role: string): boolean{
+    return this.auth.hasRealmRole(role);
+  }
+
+
+  /**
+   * Return the users realm level roles
+   */
+  public getRealmRoles():string[]{
+    if (this.auth != null && this.auth.realmAccess != null && this.auth.realmAccess.roles != null) {
```
josemigallas: Better check whether the value is falsy instead:
`if (this.auth && this.auth.realmAccess && this.auth.realmAccess.roles) ... `

josemigallas:
LGTM once lint is fixed, I think you can run `npm run lint --fix` from the root folder

ciaranRoche:
Cheers thanks @josemigallas

packages/auth/src/AuthService.ts
```diff
@@ -68,4 +68,21 @@ export class AuthService {
...
+   */
+  public hasRealmRole(role: string): boolean {
+    return this.auth.hasRealmRole(role);
+  }
+
+  /**
+   * Return the users realm level roles
+   */
+  public getRealmRoles(): string[] {
+    if (this.auth && this.auth.realmAccess && this.auth.realmAccess.roles) {
```
wtrocki: [Optional] strange to check this.auth where we do not check that in other cases.

wtrocki:
> Currently, we offer no way for a developer to check the Realm Role 

There is way. People can use `unwrap` method to get actual keycloak api and execute those actions. For this case it makes sense to expose that as this is common use case, but for the rest we may just let people to use keycloak directly.

packages/auth/src/AuthService.ts
ciaranRoche: Good catch was just an overlook on my part when trying to get it build

ciaranRoche:
@wtrocki good idea, worth adding some documentation for using `unwrap` method, I think we should still supply as many common use cases as possible.

PR #64, opened 2018-04-25
Some propositions/improvements to documentation

josemigallas:
# Description
Despite the great job @paolobueno has done, I think there are some parts that are outdated or inaccurate. So that, in this PR I am proposing a set of changes in order to improve the current docs (the majority are just re-structuring).

Some being fixes and some others improvements, they are all gathered in the following list (with a brief justification):

Fixes:
- [Fix] Some asciidoc format issues have been addressed (like numbered list)
- [Fix] In core docs, removed "Logging" as a main feature.
- [Fix] All references to "AeroGear Services" have been replaced for simply "AeroGear".

Propositions:
- [Change / Fix] The old core-cordova approach was referenced in some sections. This has been removed. I propose we leave the documentation concerning plugins for the next sprint when we implement it.
- [Change] The contributing doc has been remodeled following the format of other similar docs in github (Lerna). It has still room for improvement but the idea behind this change is to briefly and clearly describe to a new developer what do with the SDK.
- [Change] In core docs, changed `configuration` for `appConfiguration` to differentiate between service and app configuration. 
- [Change] In core docs, changed example to show only config parser (metrics will be in metrics).
- [Change] In metrics docs, extended info about usage, remodeled structure (First usage, then description)


README.md
```diff
@@ -7,12 +7,12 @@
 |                 | Project Info                                                     |
 | --------------- | ---------------------------------------------------------------- |
 | License:        | Apache License, Version 2.0                                      |
-| Build:          | Gradle                                                           |
+| Build:          | NPM                                                              |
```
josemigallas: @paolobueno @wtrocki Don't know what to put here but I assumed "Gradle" was wrong

paolobueno: Copypaste fail for me!
I guess npm is as descriptive as we can get :+1:

docs/contributing-guide.adoc
```diff
@@ -1,52 +1,121 @@
...
 [source,bash]
-----
 npm run test
-npm run lint
-----
 
-=== Using core in projects
+==== Run linter
+[source,bash]
+npm run lint
```
paolobueno: this might be missing being surrounded with `----\n`, not sure if asciidoc supports this syntax for a single source line?

docs/contributing-guide.adoc
```diff
@@ -1,52 +1,121 @@
...
+
+== Adding a new module
+
+Create a new NPM package under the packages folder with the module name: `/packages/my-module`.
+
+When creating the package.json, include the org scope to its package name: `"name": "@aerogear/my-module"`.
+
+Then add the following standard options and dependencies:
+[source.json]
+{
```
paolobueno: I think it's better to move these big code snippets to separate files, in raincatcher we created a [templates directory](https://github.com/feedhenry-raincatcher/raincatcher-core/tree/master/templates) directory to keep these.

Actually we had full example modules to copy paste into new ones, but since we needed to exclude them from lerna it was hard to use the same scripts to update dependencies, etc.

Still we can start off with moving these base package.json and tsconfig to a new dir.

docs/core.adoc
```diff
@@ -31,13 +30,14 @@ Refer to the link:./getting-started[Getting Started] guide for more information
...
 ----
-import { MetricsService } from '@aerogear/core';
-import configuration from '../mobile-services.json';
+import { ConfigurationHelper, MetricsService } from '@aerogear/core';
+import appConfiguration from '../mobile-services.json';
 
-// utilizes Core to retrieve it's own configuration
-const metrics = new MetricsService(configuration);
+const helper = new ConfigurationHelper(appConfiguration);
+const metricsConfig = helper.getConfig(MetricsService.ID);
```
paolobueno: I had replaced this with just instantiating the `MetricsService` due to https://github.com/aerogear/aerogear-js-sdk/pull/58#discussion_r183633054, because in practice the user would never need to do this.

I'm happy to keep it this way though, since it's pretty much the only thing that core does atm. If that's the case we can remove the `MetricsService` from the import at line 37

docs/getting-started.adoc
```diff
@@ -12,78 +12,16 @@ In order to develop a Cordova application utilizing the SDK you will need the fo
...
-window.MobileCore.MetricsService.getMetrics(
-  function (metrics) {
-    $("#metrics-info").html(JSON.stringify(metrics));
-  },
-  function (error) {
-    $("#metrics-info").html(error);
-  }
-);
-----
+[TODO: to be finished when we integrate plugins in the SDK (metrics)]
```
paolobueno: maybe use an asciidoc comment here: https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#comments

docs/contributing-guide.adoc
josemigallas: I've been using https://asciidoclive.com so I think syntax and format-wise there shouldn't be any problems

docs/contributing-guide.adoc
josemigallas: Awesome

docs/core.adoc
josemigallas: Ah yes, good catch.

docs/metrics.adoc
```diff
@@ -3,22 +3,39 @@
...
+metrics.publish("my-custom-type", customMetrics)
+  .then(() => {
+    console.info("Custom metrics registered successfully");
+  });
 ----
 
-== App metrics record fields
+== Default metrics
+
+The default metrics are always sent alongside any custom metrics. They are composed by the following fields:
```
paolobueno: Also referring to custom metrics here, could mention that they're sent alongside all metrics records :+1:

PR #66, opened 2018-04-27
Adds Cordova Plugin for Metrics

josemigallas:
## Motivation
To gather app and device metrics, it is necessary to install 2 plugins that provide this functionality in Cordova:
* cordova-plugin-device
* cordova-plugin-app-version

By creating a "wrapper-plugin" we:
* make our SDK independent from these plugins (they could de changed or updated over time)
* make it easier and more appealing for the end-user (only 1 plugin must be installed and it is from AeroGear)

## Description
This plugin has a JS interface that complies with Metrics SDK interfaces. That is, regardless of what the other plugins return we always ensure that our SDK doesn't break. 

## TODO
- [x] Implement and use wrapper
- [ ] Update tests

## Verification
⚠️ **Expect it to fail!**
Neither the plugin nor core@0.2.0 have been published so that they have to be installed locally.

Follow these ugly but effective steps for verification:
1. In `app/src/services/metrics.ts` copy-paste the following class:
```javascript
// import MetricsPublisher interface
import { MetricsService, MetricsPublisher } from '@aerogear/core';

// copy-paste
export class MyMetricsPublisher implements MetricsPublisher {
  public publish(payload: any): Promise<any> {
    return new Promise((res, rej) => {
      console.log(`Publishing metrics:`);
      for (const k in payload) {
        if (typeof payload[k] !== "object") {
          console.log(`${k}: ${payload[k]}`);
        } else {
          console.log(`${k}: {`);
          for (const l in payload[k]) {
            console.log(`__${l}: {`);
            for (const j in payload[k][l]) {
              console.log(`____${j}: ${payload[k][l][j]}`);
            }
            console.log(`__}`);
          }
          console.log(`}`);
        }
      }
    });
  }
}

// Lastly set it after instantiating `MetricsService`
export function initMetrics() {
  metrics = new MetricsService(appConfig);
  metrics.metricsPublisher = new MyMetricsPublisher();
  ...
```

2. Use local dependencies, by changing the following lines (adjust the paths accordingly):
```
// In package.json
"dependencies": {
  "@aerogear/core": "file:../packages/core",
}

// In config.xml
<widget>
    <plugin name="cordova-plugin-aerogear-metrics" spec="../aerogear-js-sdk/packages/cordova-plugin-aerogear-metrics" />
</widget>
```

3. Finally run the app 
```
ionic cordova emulate ios -lc
```
You should be able to see in the logs something similar to:
```
[15:36:45]  console.log: Publishing metrics:
[15:36:45]  console.log: clientId: 30121000-49f7-11e8-940e-31c8f9333461
[15:36:45]  console.log: type: init
[15:36:45]  console.log: timestamp: 1524836205048
[15:36:45]  console.log: data: {
[15:36:45]  console.log: __app: {
[15:36:45]  console.log: ____appId: io.ionic.starter
[15:36:45]  console.log: ____appVersion: 0.0.1
[15:36:45]  console.log: ____sdkVersion:
[15:36:45]  console.log: __}
[15:36:45]  console.log: __device: {
[15:36:45]  console.log: ____platform: iOS
[15:36:45]  console.log: ____platformVersion: 11.3
[15:36:45]  console.log: ____device: x86_64
[15:36:45]  console.log: __}
[15:36:45]  console.log: }
```

4. If it fails, remember to
    1. `npm run build` in `aerogear-js-sdk` before installing local dependencies.
    2. hardcore clean-install in the app:
```
rm -rf www && \\
ionic cordova plugin rm cordova-plugin-aerogear-metrics && \\
ionic cordova plugin add ../aerogear-js-sdk/packages/cordova-plugin-aerogear-metrics && \\
ionic cordova emulate ios -lc
```


packages/cordova-plugin-aerogear-metrics/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@aerogear/cordova-plugin-aerogear-metrics",
+  "version": "0.0.1",
```
wtrocki: [Info] We going to keep the versions the same. No need to change as lerna will do that.

packages/cordova-plugin-aerogear-metrics/www/metrics.js
```diff
@@ -0,0 +1,42 @@
+
+/**
+ * Get device metrics, to be called after deviceReady event
+ * 
+ * It uses cordova-plugin-device plugin.
+ * @example
+ * const deviceMetrics = await aerogear.getDeviceMetrics();
+ * @returns {Promise<DeviceMetrics>} The device metrics
+ */
+exports.getDeviceMetrics = function () {
```
wtrocki: Any reason why we have logic in the plugin? 
Could this code be in the Cordova TypeScript wrapper.
`
platform: device.platform,
platformVersion: device.version,
device: device.model
`

wtrocki: Actually it's ok to have it here.

packages/core/src/metrics/platform/CordovaAppMetrics.ts
```diff
@@ -1,15 +1,16 @@
 import { AppMetrics, Metrics } from "../model";
 
+// plugin's JS interface
+declare var aerogear: {
```
wtrocki: -1000. Plugin should declare types

wtrocki: aerogear seems unfortunate as we will not be able to declare more on this.

packages/core/src/metrics/platform/CordovaAppMetrics.ts
```diff
@@ -1,15 +1,16 @@
...
+  getAppMetrics(): Promise<AppMetrics>;
```
wtrocki: Can this be defined in plugin types rather than here.

packages/cordova-plugin-aerogear-metrics/www/metrics.js
josemigallas: That's a very good idea I didn't think of, we have already a specific Metrics implementation `CordovaAppMetrics` and `CordovaDeviceMetrics` and this logic belongs there. The JS interface is therefore not necessary.

packages/core/src/metrics/platform/CordovaAppMetrics.ts
josemigallas: What do you mean we won't be able to declare other functions? Anyway I think this declaration is no longer needed, since we can remove the JS wrapper -> (https://github.com/aerogear/aerogear-js-sdk/pull/66#discussion_r184704632)

packages/core/src/metrics/platform/CordovaAppMetrics.ts
josemigallas: You're right, this was more a workaround than anything else, anyway -> https://github.com/aerogear/aerogear-js-sdk/pull/66#discussion_r184704632

packages/core/src/metrics/platform/CordovaAppMetrics.ts
wtrocki: When having 2 declarations with the same name previous declaration will be shadowed causing some hard to find compilation errors. Been there.. :)

packages/cordova-plugin-aerogear-metrics/plugin.xml
```diff
@@ -0,0 +1,17 @@
...
+<plugin
+    id="cordova-plugin-aerogear-metrics"
+    version="0.0.1"
+    xmlns="http://apache.org/cordova/ns/plugins/1.0"
+    xmlns:android="http://schemas.android.com/apk/res/android"
+>
+    <name>cordova-plugin-aerogear-metrics</name>
+
+    <js-module name="metrics" src="www/metrics.js">
+        <clobbers target="aerogear" />
```
paolobueno: I think clobbers aerogear is a bad choice here since documentation says it'll be overwritten by other plugins, maybe ` <merges target="aerogear" />` or `<clobbers target="aerogear.metrics" />`

packages/cordova-plugin-aerogear-metrics/www/metrics.js
```diff
@@ -0,0 +1,42 @@
...
+  return new Promise(function (res, rej) {
+    res({
+      platform: device.platform,
+      platformVersion: device.version,
+      device: device.model
+    });
+  });
```
paolobueno: `return Promise.resolve({...})` for a shortcut.

packages/cordova-plugin-aerogear-metrics/plugin.xml
paolobueno: Also as @wtrocki mentioned this could also be removed and have core use the dependant plugins directly

packages/cordova-plugin-aerogear-metrics/www/metrics.js
paolobueno: +1 to make this just a bag of dependencies and have core use them directly.

packages/core/src/metrics/MetricsService.ts
```diff
@@ -72,13 +72,13 @@ export class MetricsService {
       data: {}
     };
 
-    metrics.forEach(m => {
-      payload.data[m.identifier] = m.collect();
-    });
+    for (const m of metrics) {
+      payload.data[m.identifier] = await m.collect();
+    }
```
paolobueno: Optional: this runs the promises sequentially which shouldn't be a problem, in order to run them in parallel do:
```javascript
const promises = metrics.concat(this.defaultMetrics)
  .map(m => m.collect().then(d => payload.data[m.identifier] = d));

return promises.then(() => this.publisher.publish(payload));
```

packages/cordova-plugin-aerogear-metrics/www/metrics.js
josemigallas: Good one!

packages/cordova-plugin-aerogear-metrics/plugin.xml
josemigallas: Yes 2 both, I wonder if it would be useful to leave an empty `metrics.js` with the clobbers defined, just in case in the future someone needs to use it. It could be added then of course, just saying. @paolobueno @wtrocki what do you think? Leave an scaffold or kill it?

packages/core/src/metrics/MetricsService.ts
```diff
@@ -60,7 +60,7 @@ export class MetricsService {
    * @param type type of the metrics to be published
    * @param metrics metrics instances that should be published
    */
-  public publish(type: string, metrics: Metrics[]): Promise<any> {
+  public async publish(type: string, metrics: Metrics[]): Promise<any> {
```
paolobueno: I think we could change the interface so this returns `Promise<MetricsPayload>`

packages/core/src/metrics/MetricsService.ts
josemigallas: I think I prefer the way it is, because it looks clearer, but I don't really have a strong opinion here. I remember @wtrocki was reluctant to remap the arrays though.

packages/core/src/metrics/MetricsService.ts
josemigallas: It's the MetricsPublisher who determines the return type here really so it should return something Network-related I guess?

packages/cordova-plugin-aerogear-metrics/www/metrics.js
wei-lee: +1 on just having one place define the implementation of `CordovaAppMetrics` and `CordovaDeviceMetrics`, but I am thinking if it's better to define them inside the plugin. There are 2 reasons:

1. They are cordova-specific implementation and they rely on the cordova plugins to work. From that point of, it makes sense for the code to live as part of the plugin.
2. Now the plugin is completely self-contained and it can be used on its own (although this one probably doesn't really matter as we don't have a use case to use the plugin by itself.)

That said, I think there might be some technical challenges to implement it this way - like how to invoke the platform specific implementation if they are defined inside the plugin. Not sure if it's possible, but if we can do it, I think it will make the code cleaner.

wei-lee: ok, I think that makes sense.

packages/core/src/metrics/MetricsService.ts
wtrocki: We removed loggerpublisher from other sdk. Going to do that here as well.

wtrocki:
@josemigallas is on holidays. There is very litte required to get that merged so I will take care of this and apply all changes + integrate and verify that with showcase apps.

wtrocki:
Ready for final review. Applied all changes and tested with showcase app.

Review instructions updated:

```
npm link @aerogear/core
cordova plugin add ../packages/metrics-cordova 
```

Proper logs should be displayed.

packages/core/src/metrics/MetricsService.ts
```diff
@@ -65,22 +66,27 @@ export class MetricsService {
...
-    });
+    const metricsPromise = metrics.concat(this.defaultMetrics)
+      .map(m => m.collect().then(data => {
+        payload.data[m.identifier] = data;
+      }));
 
-    this.defaultMetrics.forEach(m => {
-      payload.data[m.identifier] = m.collect();
+    return Promise.all(metricsPromise).then(() => {
+      if (this.publisher) {
```
josemigallas: This is already checked at line 69, I will remove it.

packages/core/src/metrics/MetricsService.ts
```diff
@@ -65,22 +66,27 @@ export class MetricsService {
...
       clientId: this.getClientId(),
       type,
       timestamp: new Date().getTime(),
       data: {}
     };
 
-    metrics.forEach(m => {
-      payload.data[m.identifier] = m.collect();
-    });
+    const metricsPromise = metrics.concat(this.defaultMetrics)
```
josemigallas: 80-83 is a little bit convoluted, do we prefer this format instead of the `async/await`?

packages/core/src/metrics/MetricsService.ts
```diff
@@ -65,22 +66,27 @@ export class MetricsService {
       throw new Error(`Type is invalid: ${type}`);
     }
 
+    if (!this.publisher || !this.defaultMetrics) {
+      return Promise.resolve();
```
josemigallas: Don't we want to throw an error in this case?

josemigallas: Also, why defaultMetrics are not always set by default? We can skip the check by calling `buildDefaultMetrics` always in the constructor.

wtrocki: This is new requirements that are already implemented on android and IOS. Basically when config is not there there should be no errors. Thats classical scenario.

packages/core/src/metrics/MetricsService.ts
josemigallas: I see, I read the recommended approach is to encapsulate it in a const `const publisher = this.publisher` to avoid nested checks.

packages/core/src/metrics/MetricsService.ts
josemigallas: Fair enough

packages/core/src/metrics/MetricsService.ts
josemigallas: The problem I see here is that you could be doing `metricsService.sendAppAndDevice()` and user would not get any feedback.

On the other hand, I think default metrics should be instantiated in constructor and never be undefined.

wtrocki: If metrics is disabled then there is no need to collect or instantiate default metrics. No strong opinions about it. I just followed how I implemented that on IOS but there could be better way

wtrocki: Ah.. get you now. As per recent req users should not interact with metrics. This metod is called on constructor

wtrocki: Ah.. get you now. As per recent req users should not interact with metrics. This metod is called on constructor

wtrocki: Ah.. get you now. As per recent req users should not interact with metrics. This metod is called on constructor

paolobueno: We could a `console.warn("Metrics service disabled due to lack of configuration")` or something, no failures but a runtime warning.

wtrocki: Yeah similar message is above this

PR #67, opened 2018-05-01
added getToken

StephenCoady:
## Motivation

We need to be able to retrieve the token using the SDK.
 
## Description
Added method to retrieve token.

## Verification
This ticket can be verified using [this](https://github.com/aerogear/cordova-showcase-template/pull/24) PR from the showcase application. See that PR for steps.

### DIFF
--- a/packages/auth/src/AuthService.ts
+++ b/packages/auth/src/AuthService.ts
@@ -97,4 +97,11 @@ export class AuthService {
     }
     return [];
   }
+
+  /**
+   * Return the access token
+   */
+  public getToken(): string {
+    return this.auth.token || "{}";
+  }
 }


wtrocki:
Happy to approve it but...

`getToken` is method where on keycloak we dealing with field. Users can use unwrap method, but make we can rename it to general method that could be used to get access to keycloak object and also set of the properties. 

If we want to go this way then let's make this property so our wrapper will be *compatibile* with keycloak solutions available on stack overflow etc. `getToken` is more java like implementation.

StephenCoady:
closing this PR as @wtrocki has rightly pointed out there is no point in providing a method to expose something we already have access to. using extract() now instead.

PR #68, opened 2018-05-02
fix: add configurable logger to SDK

wtrocki:
## Motivation

Add option to kill logging or extend it. We could add more debug stuff after and disable it for end users.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1,6 +1,8 @@
 /**
  * @module @aerogearservices/core
  */
+import console from "loglevel";
+console.setDefaultLevel(console.levels.WARN);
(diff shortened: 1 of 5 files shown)


packages/core/src/index.ts
```diff
@@ -1,6 +1,8 @@
 /**
  * @module @aerogearservices/core
  */
+import console from "loglevel";
+console.setDefaultLevel(console.levels.WARN);
```
paolobueno: Maybe INFO?

wtrocki: -1. This is default for users and info should not appear by default in the browser, cordova etc. We can change it.

PR #69, opened 2018-05-02
User api experiment

wtrocki:
## Motivation

I was thinking how we can satisfy end user api requirements and also metrics auto init. Our requirements are:

1) Allow each SDK to be fully independent
2) Allow us to maintain configuration and SDK initialization inside SDK instead of users being responsible for that.
3) [Technical] Avoid circular dependencies (auth depending on core and core on auth)
4) Allow full web support when we compile packages to plain bundled js code for independent libraries.

Happy for us there are many examples for such usage. 
For example: https://firebase.google.com/docs/web/setup

TL;DR - we need new package that will be aggregating all SDK's and making sure that they will receive config.

## IDEA 

Users will just do
```
<script src="aerogear-app.js"></script>
// common.js version
var app = require("@aerogear/app")
app.initialize(config);
let auth = app.getAuth();
```

However having it this way (bundle all in app) will jeopardize point 1) of having independent packages. In this PR I'm going to spike how we can get all the goodness and flexible user api without jeopardizing all the points. I will also try to provide early support for the web to spike how this could work.

This idea is also going to resolve couple problems:
- autoconfiguration of the push plugin variants.
- making metrics work outside the box. Currently it happens in user space which is not desired: https://github.com/aerogear/cordova-showcase-template/blob/master/src/services/metrics.ts

Note: This is the raw idea that needs to be spiked for all options. 
See commit: 863408e73a0c1f4ab9d789a9daef28da4d5a8d96

Concerns: this change will affect separation for SDK's as they going to be exposed in global window objects.

### DIFF
--- a/packages/app/src/index.ts
+++ b/packages/app/src/index.ts
@@ -1,0 +1,17 @@
+import { AeroGearConfiguration, ConfigurationHelper } from "@aerogear/core";
+
+declare var window: any;
+
+let config: ConfigurationHelper;
+
+export function init(configuration: AeroGearConfiguration) {
+  config = new ConfigurationHelper(configuration);
+  window.Metrics.init(config);
+}
+
+export function getAuth() {
+  if (config && window.Auth) {
+    window.Auth.init(config);
+    return window.Auth;
+  }
+}
--- a/packages/app/.npmignore
+++ b/packages/app/.npmignore
@@ -1,0 +1,4 @@
+test/
+coverage/
+.nyc_output/
+node_modules/
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -1,0 +1,73 @@
+{
+  "name": "@aerogear/app",
+  "version": "0.1.0",
+  "description": "AeroGear Services main module used for configuration purposes",
+  "main": "dist/index.js",
+  "types": "types/index.d.ts",
+  "author": "Aerogear <aerogear@googlegroups.com>",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/aerogear/aerogear-js-sdk.git"
+  },
+  "bugs": {
+    "url": "https://issues.jboss.org/projects/AGCORDOVA/issues"
+  },
+  "keywords": [
+    "aerogear",
+    "mobile",
+    "keycloak",
+    "sso",
+    "auth",
+    "openid"
+  ],
+  "license": "Apache-2.0",
+  "scripts": {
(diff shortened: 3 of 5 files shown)


packages/app/src/index.ts
```diff
@@ -0,0 +1,17 @@
...
+declare var window: any;
+
+let config: ConfigurationHelper;
+
+export function init(configuration: AeroGearConfiguration) {
+  config = new ConfigurationHelper(configuration);
+  window.Metrics.init(config);
+}
+
+export function getAuth() {
```
wtrocki: Going to check if we can still reference actual types using `///` notation so users will benefit from types.

wtrocki:
@wei-lee This is really raw idea that I tried to make work in order to spike top level API. I not going to use window. 

I'm more interested about your opinions of top level api:
```
<script src="aerogear-app.js"></script>
// common.js version
var app = require("@aerogear/app")
app.initialize(config);
let auth = app.getAuth();
```

WDYT about this?

josemigallas:
@wtrocki In my opinion your snippet is a good idea, it is what I would expect as user. I would expect however that this "app" package is the Core. Core should be the one thing that binds every other service SDK.

About the circular dependencies, I think we can simply use NPM's "peerDependency".

wei-lee:
@wtrocki so that that means all the modules (including core and service modules) will be bundled as part of `aerogear-app`?

wtrocki:
So the story is. 

I was thinking how we can use those in core without linking them directly to core and have circular dependency. I could not get that running without crazy hacks. So borrowed idea from firebase where they use global variables and global app package.
They still have core package as well. I do not think that window is bad. 
For classical web support we will need to have some global namespace unless we want to support it only by providing common.js packages.

As for app itself I'm currently investigating options we can do it. 
Simplest way to avoid globals be to have something like this on user side:

```
var app = require("@aerogear/app") 
app.auth(require("@aerogear/auth"));

app.init(config);
let auth = app.getAuthSdk();
```

wtrocki:
Update from offline discussion:

- We going to try to use core package for init
- Core should be using checked fixed location (to be investigated for each platform)

wei-lee:
> Core should be using checked fixed location (to be investigated for each platform)

can you explain a bit more about it?

wtrocki:
@wei-lee Instead of users providing configuration object in their application we will require this file to be placed in www folder and core will read that automatically. We will still provide way to override this, but it will be configured from expected location as on Android and IOS.

PR #70, opened 2018-05-03
fix: use type for config

wtrocki:
## Motivation

Use type for fetching mobile configurations instead of name.
Name is just human readable name and it's not guaranteed to be unique.

See: https://github.com/aerogear/aerogear-android-sdk/pull/217

### DIFF
--- a/packages/core/src/configuration/ConfigurationHelper.ts
+++ b/packages/core/src/configuration/ConfigurationHelper.ts
@@ -17,10 +17,10 @@ export class ConfigurationHelper {
 
   /**
    * Get a service configuration object, provided an existing id is given
-   * @param id string - The id of the service
+   * @param type string - The type of the service
    */
-  public getConfig(id: string): ServiceConfiguration | undefined {
-    return find(this.configurations, service => service.id.toLowerCase() === id.toLowerCase());
+  public getConfig(type: string): ServiceConfiguration | undefined {
+    return find(this.configurations, service => service.type.toLowerCase() === type.toLowerCase());
   }
 
 }
--- a/packages/core/src/configuration/ConfigurationHelper.ts
+++ b/packages/core/src/configuration/ConfigurationHelper.ts
@@ -16,11 +16,11 @@ export class ConfigurationHelper {
   }
 
   /**
-   * Get a service configuration object, provided an existing id is given
+   * Get a service configuration object, provided an existing type is given
    * @param type - The type of the service
    */
-  public getConfig(type: string): ServiceConfiguration | undefined {
-    return find(this.configurations, service => service.type.toLowerCase() === type.toLowerCase());
+  public getConfigByType(type: string): ServiceConfiguration[] | undefined {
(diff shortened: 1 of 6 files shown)


josemigallas:
@wtrocki But `type` is the fixed name like "metrics" and "auth". The `id` is the name of the custom service like "my-metrics-service", different than the actual `name`. Am I wrong? Type is what we had and we changed it precisely to support custom ids.

packages/core/src/configuration/ConfigurationHelper.ts
```diff
@@ -17,10 +17,18 @@ export class ConfigurationHelper {
 
   /**
    * Get a service configuration object, provided an existing id is given
-   * @param type string - The type of the service
+   * @param type - The type of the service
    */
   public getConfig(type: string): ServiceConfiguration | undefined {
```
wei-lee: should this method actually return an array?

packages/core/src/configuration/ConfigurationHelper.ts
```diff
@@ -17,10 +17,18 @@ export class ConfigurationHelper {
...
+  public getConfig(type: string): ServiceConfiguration | undefined {
+    return find(this.configurations, service => service.type.toLowerCase() === type.toLowerCase());
+  }
+
+  /**
+   * Get a service configuration object, provided an existing id is given
+   * @param id - unique id of the service
+   */
+  public getConfigById(id: string): ServiceConfiguration | undefined {
+    return find(this.configurations, service => service.type.toLowerCase() === id.toLowerCase());
```
danielpassos: `service.type` should not be `service.id`?

packages/core/src/configuration/ConfigurationHelper.ts
danielpassos: Very good point!

josemigallas: This method filters the `services` array and returns only one ServiceConfig, ready to be used. Is there any reason to return an array instead?

packages/core/src/configuration/ConfigurationHelper.ts
josemigallas: `id` is generated by OpenShift so that it is not fixes, type on the other hand will be always `metrics`, `auth`, etc.

wei-lee: but this method is called `getConfigById` and it should perform the look up by id? Look up by `type` is done in the method above.

packages/core/src/configuration/ConfigurationHelper.ts
wei-lee: yes, there could be multiple `custom-runtime-connector` services and they all will have the same types, but each with different id.

josemigallas: I don't understand why we get the first one then (`[0]`) if there could be many of them, I think we should still be filtering by `id`.

wtrocki: @josemigallas see conversation on the list for reasoning. Happy to explain that as well.

PR #72, opened 2018-05-04
[Do not merge] AEROGEAR-2527 Spike ensuring device check approach

aidenkeating:
## Motivation

Just like in the other SDKs we need device checks.

## Description

Ensure that the approach of exposing other plugins within our SDK would work.

Note that this is a very simple approach just to confirm that everything works together, the final approach will be structured similar to the Android SDK checks.

### DIFF
--- a/packages/security/src/index.ts
+++ b/packages/security/src/index.ts
@@ -1,0 +1,7 @@
+declare var IRoot: any;
+
+export default class SecurityService {
+  constructor() {}
+
+  public isRooted = IRoot.isRooted;
+}
--- a/packages/security-cordova/package.json
+++ b/packages/security-cordova/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "@aerogear/cordova-plugin-aerogear-security",
+  "version": "0.1.0",
+  "description": "Cordova plugin for AeroGear Security",
+  "cordova": {
+    "id": "cordova-plugin-aerogear-security",
+    "platforms": []
+  },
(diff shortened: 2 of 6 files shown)


aidenkeating:
This will more than likely be replaced with an actual PR for rooted checks etc. This was the output of a spike to confirm the approach.

josemigallas:
@aidenkeating Looks good and simple. I would only recommend consistency in names. IMO package (prefixed with @aerogear) and plugin should be the same.

@wtrocki we could add a new method in `MetricsService` that checks whether or no `IRoot` exists, something like `sendDeviceCheckMetrics`? Or some options passed in the constructor:
```
new MetricsService({ deviceCheck: true });
```

wtrocki:
@josemigallas This should be initiated by other SDK created in this PR. Core should now know anything about security metrics.

packages/security/src/index.ts
```diff
@@ -0,0 +1,7 @@
+declare var IRoot: any;
+
+export default class SecurityService {
+  constructor() {}
+
+  public isRooted = IRoot.isRooted;
```
wei-lee: is it possible that `IRoot` is undefined?

wtrocki: Yeah. This code should use the same aproach as in core where we:

1) Detect platform
2) Detect if plugin is available.

aidenkeating:
Closing this to be replaced with other, proper, PRs

PR #73, opened 2018-05-04
Agcordova 224

wtrocki:
## Motivation

Update docs to reflect that there is no end user api.


docs/metrics.adoc
```diff
@@ -1,23 +1,9 @@
-= AeroGear Services Metrics SDK
+= AeroGear Services Metrics documentation
 
-Metrics is another service provided by Core and no further dependencies are required for its utilization.
-It provides registration of default metrics for the application such as App and Device information as well as supplying these services to other modules of the SDK.
+This document explains how to use the AeroGear Metrics SDK and integrate it with an App.
+Metrics are available automatically when metrics service is available in `mobile-services.json` configration.
```
josemigallas: Maybe mention that it sends metrics automatically under "usage" section? Just to be explicit about it.

PR #77, opened 2018-05-08
AEROGEAR-2527 Initial approach for device checks SDK

aidenkeating:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-2674

## Description
Add a way to run/create own tests based on the Android SDK.

## Progress
- [x] Main stuff
- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/security-cordova/package.json
+++ b/packages/security-cordova/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "@aerogear/cordova-plugin-aerogear-security",
+  "version": "0.1.0",
+  "description": "Cordova plugin for AeroGear Security",
+  "cordova": {
+    "id": "cordova-plugin-aerogear-security",
+    "platforms": []
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/aerogear/aerogear-js-sdk.git"
+  },
+  "keywords": [
+    "ecosystem:cordova"
+  ],
+  "author": "AeroGear <aerogear@googlegroups.com>",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/aerogear/aerogear-js-sdk/issues"
+  },
+  "homepage": "https://github.com/aerogear/aerogear-js-sdk#readme"
+}
--- a/packages/security-cordova/package.json
+++ b/packages/security-cordova/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@aerogear/cordova-plugin-aerogear-security",
   "version": "0.1.0",
-  "description": "Cordova plugin for AeroGear Security",
+  "description": "Cordova plugin for self defence checks for Android and iOS devices.",
   "cordova": {
     "id": "cordova-plugin-aerogear-security",
     "platforms": []
--- a/packages/security-cordova/plugin.xml
+++ b/packages/security-cordova/plugin.xml
@@ -1,0 +1,12 @@
+<?xml version='1.0' encoding='utf-8'?>
+<plugin
+    id="cordova-plugin-aerogear-device-trust"
+    version="0.0.1"
+    xmlns="http://apache.org/cordova/ns/plugins/1.0"
+    xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <name>cordova-plugin-aerogear-device-trust</name>
+
+    <dependency id="cordova-plugin-iroot" version="0.7.0" />
+
+</plugin>
--- a/packages/security-cordova/plugin.xml
+++ b/packages/security-cordova/plugin.xml
@@ -1,11 +1,11 @@
 <?xml version='1.0' encoding='utf-8'?>
 <plugin
-    id="cordova-plugin-aerogear-device-trust"
-    version="0.0.1"
+    id="cordova-plugin-aerogear-security"
+    version="0.1.0"
     xmlns="http://apache.org/cordova/ns/plugins/1.0"
     xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <name>cordova-plugin-aerogear-device-trust</name>
+    <name>cordova-plugin-aerogear-security</name>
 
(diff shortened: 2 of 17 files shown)


packages/security-cordova/plugin.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version='1.0' encoding='utf-8'?>
+<plugin
+    id="cordova-plugin-aerogear-device-trust"
+    version="0.0.1"
```
josemigallas: This should be the same version as in `package.json`.

packages/security-cordova/plugin.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version='1.0' encoding='utf-8'?>
+<plugin
+    id="cordova-plugin-aerogear-device-trust"
```
josemigallas: I don't know if this is a requirement but I think we should name plugin and package the same.

aidenkeating: Nice catch, thanks

aidenkeating:
@wtrocki @wei-lee Mind taking a look?

packages/security/src/SecurityService.ts
```diff
@@ -0,0 +1,7 @@
+import { SecurityCheck, SecurityCheckResult } from "./deviceTrust";
+
+export class SecurityService {
+  public check(check: SecurityCheck): Promise<SecurityCheckResult> {
+    return check.test();
+  }
```
wei-lee: I think we should also support run an list of checks

packages/security-cordova/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@aerogear/cordova-plugin-aerogear-security",
+  "version": "0.1.0",
+  "description": "Cordova plugin for AeroGear Security",
```
wtrocki: This is overloaded term. Best to say what `AeroGear Security` means?

packages/security/src/deviceTrust/SecurityCheckType.ts
```diff
@@ -0,0 +1,7 @@
+import { RootedCheck } from "./checks";
+
+export class SecurityCheckType {
```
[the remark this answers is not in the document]
aidenkeating: Yeah documentation is still in TODO (see description of issue).

This is loosely following https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckType.java. However, I didn't think TypeScript had support for expanded enums like Java has, thought they were number-based only? Hence the class

packages/security/src/deviceTrust/checks/RootedCheck.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  public test(): Promise<SecurityCheckResult> {
+    return new Promise((resolve, reject) => {
+      if (!IRoot) {
+        reject(new Error("Could not find plugin IRoot."));
+        return;
+      }
+
+      IRoot.isRooted((rooted: number) => {
+        const result = new SecurityCheckResult(this.name, !!rooted);
```
wtrocki: That's strange in TS. You usually do something like:

```
const result: SecurityCheckResult = { name, rooted }
```

wtrocki:
Got that running locally. We are missing README for plugin. That is somehow recommended to cordova plugins.

packages/security/src/deviceTrust/SecurityCheckType.ts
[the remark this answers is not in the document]
wtrocki: I meant that I do not understand why we need this class as it has just one static member.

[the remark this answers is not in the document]
aidenkeating: Oh this will be expanded with the rest of the inbuilt checks as they're added

packages/security/src/deviceTrust/checks/RootedCheck.ts
aidenkeating: That makes sense, will address this

packages/security/src/SecurityService.ts
aidenkeating: @wei-lee There is another ticket to address the list of checks properly. Mainly to make this PR as quick as possible to unblock all other tasks in the story.

aidenkeating:
@wtrocki Mind taking another look please? Have created some tasks to address some of the larger issues. Would like to unblock the rest of the tasks in the story if possible though.

packages/security/test/SecurityService.ts
```diff
@@ -0,0 +1,35 @@
...
+
+describe("SecurityService", () => {
+
+  beforeEach(() => {
+    securityService = new SecurityService();
+  });
+
+  describe("#test", () => {
+
+    it("should pass", (done) => {
```
josemigallas: I would extend the tests name more. Each `it` should describe a test case, not the result. _It should pass if the check passes_ for example.

aidenkeating:
@josemigallas @wtrocki Mind taking a look?

packages/security/src/deviceTrust/SecurityCheckType.ts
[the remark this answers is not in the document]
wtrocki: Ah.. Then you probably want to do something like 

export let checks = [new RootedCheck()];

wtrocki:
Merging. Additional comments can be handled in separate PR.

packages/security-cordova/plugin.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version='1.0' encoding='utf-8'?>
+<plugin
+    id="cordova-plugin-aerogear-security"
+    version="0.1.0"
```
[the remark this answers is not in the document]
josemigallas: I meant whenever the version is bumped in the plugin's package.json, since it won't be updated automatically by lerna. For this particular case, both are `0.1.0` so all good, sorry for late response!

PR #78, opened 2018-05-08
fix: run metrics only for native platform

wtrocki:
## Motivation

Run metrics only for native implementations. 
Currently only Cordova.


packages/core/src/PlatformUtils.ts
```diff
@@ -14,3 +14,10 @@ export const isMobileCordova = () => (
   // tslint:disable-next-line:no-string-literal
   !!(window["cordova"] || window["phonegap"] || window["PhoneGap"])
 );
+
+/**
+ * Detect if device is running in native environment as opposite to browser platform
+ *
+ * @return {boolean} Is running in native
+ */
+export const isNative = () => isMobileCordova();
```
paolobueno: :+1: nice alias since it can be turned into `isMobileCordova() || isMobileReactNative()`, etc

PR #79, opened 2018-05-08
AEROGEAR-2672 Implement debug security check

ciaranRoche:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-2676

## Description
Currently, we do not have a check to see if a device is running in debug mode. This PR adds a debug check by using a cordova plugin which is added to our aerogear security plugin

## Verify
The following PR can be used to verify these changes: https://github.com/aerogear/cordova-showcase-template/pull/29

In SDK root folder run:
```
npm install
npm run bootstrap
npm run build
```
cd to `cordova-showcase` and checkout above PR and run the following:
```
cordova plugin add --link <<path to security-cordova>>
npm install
npm run ionic:<<ios or android>>
```

### DIFF
--- a/packages/security/src/deviceTrust/SecurityCheckType.ts
+++ b/packages/security/src/deviceTrust/SecurityCheckType.ts
@@ -1,4 +1,4 @@
-import { NonEmulatedCheck, NonRootedCheck, NonDebugCheck } from "./checks";
+import { NonDebugCheck, NonEmulatedCheck, NonRootedCheck } from "./checks";
 
 /**
  * Detect whether a device is rooted (Android) or Jailbroken (iOS).
@@ -12,6 +12,3 @@ export const notEmulated = new NonEmulatedCheck();
    * Detect whether a device is running in debug mode.
    */
 export const notDebugMode = new NonDebugCheck();
-
-
-
--- a/packages/security/src/deviceTrust/SecurityCheckType.ts
+++ b/packages/security/src/deviceTrust/SecurityCheckType.ts
@@ -1,4 +1,4 @@
-import { NonEmulatedCheck, NonRootedCheck } from "./checks";
+import { NonEmulatedCheck, NonRootedCheck, DebuggerCheck } from "./checks";
 
 /**
  * Detect whether a device is rooted (Android) or Jailbroken (iOS).
@@ -8,3 +8,5 @@ export const notRooted = new NonRootedCheck();
  * Detect whether a device is running on an emulator
  */
 export const notEmulated = new NonEmulatedCheck();
+
+export const notDebugger = new DebuggerCheck();
--- a/packages/security/src/deviceTrust/SecurityCheckType.ts
+++ b/packages/security/src/deviceTrust/SecurityCheckType.ts
@@ -1,4 +1,4 @@
-import { NonEmulatedCheck, NonRootedCheck, DebuggerCheck } from "./checks";
+import { NonEmulatedCheck, NonRootedCheck, NonDebugCheck } from "./checks";
 
 /**
  * Detect whether a device is rooted (Android) or Jailbroken (iOS).
@@ -9,9 +9,9 @@ export const notRooted = new NonRootedCheck();
  */
 export const notEmulated = new NonEmulatedCheck();
   /**
-   * A check for whether a debugger is attached to the current application
+   * Detect whether a device is running in debug mode.
    */
-export const IsDebuggerConnected = new DebuggerCheck();
+export const notDebugMode = new NonDebugCheck();
 
 
 
--- a/packages/security/src/deviceTrust/SecurityCheckType.ts
+++ b/packages/security/src/deviceTrust/SecurityCheckType.ts
@@ -8,5 +8,10 @@ export const notRooted = new NonRootedCheck();
  * Detect whether a device is running on an emulator
  */
 export const notEmulated = new NonEmulatedCheck();
+  /**
+   * A check for whether a debugger is attached to the current application
+   */
+export const IsDebuggerConnected = new DebuggerCheck();
 
-export const notDebugger = new DebuggerCheck();
+
+
--- a/packages/security/src/deviceTrust/checks/DebuggerCheck.ts
+++ b/packages/security/src/deviceTrust/checks/DebuggerCheck.ts
@@ -3,18 +3,24 @@ import { SecurityCheckResult } from "../SecurityCheckResult";
 
 declare let cordova: any;
 
+/**
+ * A check for whether a debugger is attached to the current application
+ */
 export class DebuggerCheck implements SecurityCheck {
     public readonly name = "Is Debugger Check";
 
+    /**
+     * An application running with an attached debugger can have its internals exposed.
+     */
     public check(): Promise<SecurityCheckResult> {
         return new Promise((resolve, reject) => {
-            if (!cordova.plugins.isDebug) {
+            if (!cordova.plugins.IsDebug) {
                 reject(new Error("Could not find plugin isDebug."));
                 return;
             }
 
-            cordova.plugins.isDebug((passed: boolean) => {
-                const result: SecurityCheckResult = {name: this.name, passed};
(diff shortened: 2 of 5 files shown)


packages/security/src/deviceTrust/checks/DebuggerCheck.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    public check(): Promise<SecurityCheckResult> {
+        return new Promise((resolve, reject) => {
+            if (!cordova.plugins.isDebug) {
+                reject(new Error("Could not find plugin isDebug."));
+                return;
+            }
+
+            cordova.plugins.isDebug((passed: boolean) => {
+                const result: SecurityCheckResult = {name: this.name, passed};
```
aidenkeating: Based on this description of the result interface in the Android SDK could we make this so that the more secure result is a pass, and the less secure result is a fail? That way all of our SDKs will be the same in their approach.

https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckResult.java#L24

I will be updating the rooted check to do the same. Updating it to be `NotRootedCheck`

See (https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/main/java/org/aerogear/mobile/security/checks/NoDebuggerCheck.java) for the debugger check.

packages/security/src/deviceTrust/SecurityCheckType.ts
```diff
@@ -1,7 +1,9 @@
 import { RootedCheck } from "./checks";
+import { DebuggerCheck } from "./checks";
 
 export class SecurityCheckType {
   public static readonly IsRooted = new RootedCheck();
+  public static readonly IsDebugger = new DebuggerCheck();
```
josemigallas: Sorry for the fastidiousness but isn't **debugger** a bit misleading? What about `isDebugMode` or simply `isDebug`. Debugger somewhat suggest that this detects if the debugger is being used while it is only detecting wether the app is on debug or release mode.

packages/security/src/deviceTrust/checks/DebuggerCheck.ts
```diff
@@ -0,0 +1,22 @@
+import { SecurityCheck } from "../SecurityCheck";
+import { SecurityCheckResult } from "../SecurityCheckResult";
+
+declare let cordova: any;
+
+export class DebuggerCheck implements SecurityCheck {
+    public readonly name = "Is Debugger Check";
```
josemigallas: static? And for constants names please use UPPER_CASE.

packages/security/src/deviceTrust/checks/DebuggerCheck.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    public check(): Promise<SecurityCheckResult> {
+        return new Promise((resolve, reject) => {
+            if (!cordova.plugins.isDebug) {
```
josemigallas: `cordova.plugin` would be `undefined` if not in Cordova (aka browser) should we consider this @wtrocki ? I think it's the same in the other Security check.

packages/security/src/deviceTrust/SecurityCheckType.ts
TommyJ1994: `isDebuggerConnected` might be more accurate if we are following what we used in the Android SDK - https://developer.android.com/reference/android/os/Debug#isdebuggerconnected

packages/security/src/deviceTrust/checks/DebuggerCheck.ts
wtrocki: -1. That's Java land. Cannot find any example for this in ts[1]. Let's not interfere with style until we have strict style guide. In RainCatcher we were using[2]

[1] https://basarat.gitbooks.io/typescript/docs/const.html
[2] https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines

packages/security/src/deviceTrust/checks/DebuggerCheck.ts
wtrocki: It's unfortunate because plugin itself is using ancient namespace: https://github.com/mattlewis92/cordova-plugin-is-debug/blob/master/plugin.xml#L8
We need to check all starting from `window.cordova`

packages/security/src/deviceTrust/SecurityCheckType.ts
josemigallas: But what are we detecting exactly? The attached debugger or the build-config called "debug"? (aka development mode)

TommyJ1994: @josemigallas A connected java debugger. The build config debug flag is under `ApplicationInfo.FLAG_DEBUGGABLE`.

TommyJ1994: But looks like the plugin we are using only checks for that flag - https://github.com/mattlewis92/cordova-plugin-is-debug/blob/master/src/android/IsDebug.java#L34 

On the other hand, the Android/iOS templates/SDK's actually check if a debugger is attached.

wtrocki: Those are two completely different things. Debug means that you can use for example self signed certs etc. IMHO Checking if debugger is attached has not much value.

packages/security/src/deviceTrust/checks/DebuggerCheck.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export class DebuggerCheck implements SecurityCheck {
+    public readonly name = "Is Debugger Check";
+
+    /**
+     * An application running with an attached debugger can have its internals exposed.
+     */
+    public check(): Promise<SecurityCheckResult> {
+        return new Promise((resolve, reject) => {
+            if (!cordova.plugins.IsDebug) {
```
wtrocki: > undefined has no field plugins 

:)

packages/security/src/deviceTrust/SecurityCheckType.ts
aidenkeating: @TommyJ1994 Do you think the current plugin we're using here does what we need? Or should we be looking for a new plugin/create our own.

packages/security/src/deviceTrust/checks/DebuggerCheck.ts
wtrocki: > undefined has no field plugins 

:) We need to check all as this plugin mounts into nested namespace.

packages/security/src/deviceTrust/SecurityCheckType.ts
wtrocki: @aidenkeating Yeah - This one was actually created for detecting that case - checking if you do not send debug version somewhere where you should not.

TommyJ1994: Well we are trying to defend against the *action* of a malicious user debugging as per OWASP/NowSecure guidelines. We could do both, don't care or just make a note, but was just giving a heads up that the cordova implementation is different to android/iOS currently.

wtrocki: Actually maybe we can just open conversation on the list as it will be quite easy to get both done, but we need to get consensus if we need both or at least be consistent in naming and name this one differently so it will appear in different category in grafana metrics.

StephenCoady:
needs rebase for verification

packages/security/src/deviceTrust/checks/NonDebugCheck.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * A check for whether a device is running in debug mode
+ */
+export class NonDebugCheck implements SecurityCheck {
+    public readonly name = "Is Debugger Check";
+
+    public check(): Promise<SecurityCheckResult> {
+        return new Promise((resolve, reject) => {
+            if (!cordova || !cordova.plugins || !cordova.plugins.IsDebug) {
```
wtrocki: ❤️

PR #81, opened 2018-05-09
Add documentation and refactor check type

aidenkeating:
## Motivation

Currently there is no API documentation for the security service.

## Description
Add API documentation for the security service.

There is also a small refactor of the `SecurityCheckType` module.

## Progress
- [x] API docs
- [x] Refactor SecurityCheckType
- [x] Make isRooted check consistent with the Android SDK implementation.

### DIFF
--- a/packages/security/src/SecurityService.ts
+++ b/packages/security/src/SecurityService.ts
@@ -1,6 +1,15 @@
 import { SecurityCheck, SecurityCheckResult } from "./deviceTrust";
 
+/**
+ * Service module for handling performing and reporting checks.
+ */
 export class SecurityService {
+
+  /**
+   * Execute the provided security check and return the result.
+   *
+   * @returns {Promise<SecurityCheckResult>}
+   */
   public check(check: SecurityCheck): Promise<SecurityCheckResult> {
     return check.check();
   }
--- a/packages/security/src/SecurityService.ts
+++ b/packages/security/src/SecurityService.ts
@@ -1,14 +1,18 @@
 import { SecurityCheck, SecurityCheckResult } from "./deviceTrust";
 
 /**
- * Service module for handling performing and reporting checks.
+ * Service module for handling performing and reporting  possible security
+ * issues in a mobile application.
+ *
+ * This requires the @aerogear/cordova-plugin-aerogear-security plugin to be
+ * included in an application.
  */
 export class SecurityService {
 
   /**
    * Execute the provided security check and return the result.
    *
-   * @returns {Promise<SecurityCheckResult>}
+   * @returns The result of the provided check.
    */
   public check(check: SecurityCheck): Promise<SecurityCheckResult> {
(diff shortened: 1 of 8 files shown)


packages/security/src/SecurityService.ts
```diff
@@ -1,6 +1,15 @@
 import { SecurityCheck, SecurityCheckResult } from "./deviceTrust";
 
+/**
+ * Service module for handling performing and reporting checks.
```
wtrocki: handling**,** performing and reporting possible security issues in mobile application. 

Worth to add info that separate plugin is required.

packages/security/src/deviceTrust/SecurityCheck.ts
```diff
@@ -1,7 +1,18 @@
 import { SecurityCheckResult } from "./SecurityCheckResult";
 
+/**
+ * Interface for a device security single check that can be executed.
```
wtrocki: security single check => single security check

packages/security/src/deviceTrust/SecurityCheck.ts
```diff
@@ -1,7 +1,18 @@
...
 export interface SecurityCheck {
+  /**
+   * The name of the check. This can be used for reporting the checks results.
+   */
   name: string;
 
+  /**
+   * Execute the check.
+   *
+   * @returns {Promise<SecurityCheckResult>}
```
wtrocki: TSdocs actually generates this types so this is redundant. We can just say in normal words what is returned.

packages/security/src/deviceTrust/SecurityCheckResult.ts
```diff
@@ -1,4 +1,19 @@
...
+ * Interface for the results of a single pass/fail security check.
+ */
 export interface SecurityCheckResult {
+  /**
+   * The name of the check. This can be used for reporting the check results.
+   */
   name: string;
+
+  /**
+   * Whether the check passed or failed. A successful check means that the environment this
```
wtrocki:
```
Whether the check passed or failed. A successful check means that the environment this application is running in is more secure than otherwise, as opposed to signalling if a certain feature was enabled
```
How about:

Flag that determines if executed check finished with success or failure.

aidenkeating:
@wtrocki Thanks for the quick review! Really appreciated! 😄 I think I addressed everything.

Would you mind taking a look at the commit I've added (https://github.com/aerogear/aerogear-js-sdk/pull/81/commits/18c373d8a7112fcaf5d548d28414919c2b84c217). It makes the rooted check consistent with the Android SDK in name and functionality (it fails when the device is rooted).

https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/main/java/org/aerogear/mobile/security/checks/NonRootedCheck.java

wtrocki:
@aidenkeating Checked. Really good job!

PR #82, opened 2018-05-09
AEROGEAR-2680 - add emulator security check

rachael-oregan:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2680

## Description
Added the emulator security check

### DIFF
--- a/packages/security-cordova/plugin.xml
+++ b/packages/security-cordova/plugin.xml
@@ -8,5 +8,6 @@
     <name>cordova-plugin-aerogear-security</name>
 
     <dependency id="cordova-plugin-iroot" version="0.7.0" />
+    <dependency id="cordova-plugin-device" version="2.0.2">
 
 </plugin>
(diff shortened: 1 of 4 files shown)


packages/security/src/deviceTrust/checks/EmulatorCheck.ts
```diff
@@ -0,0 +1,19 @@
+import { SecurityCheck } from "../SecurityCheck";
+import { SecurityCheckResult } from "../SecurityCheckResult";
+
+declare var device: any;
+
+export class EmulatorCheck implements SecurityCheck {
```
aidenkeating: This needs some documentation

packages/security-cordova/plugin.xml
```diff
@@ -8,5 +8,6 @@
     <name>cordova-plugin-aerogear-security</name>
 
     <dependency id="cordova-plugin-iroot" version="0.7.0" />
+    <dependency id="cordova-plugin-device" version="2.0.2">
```
wtrocki: I will need to test it as we already using this version of the plugin in the metrics-cordova. Not sure if having both will not clash or cause some issues.

PR #84, opened 2018-05-09
feat: Add script to update plugin.xml on release

paolobueno:
## Motivation

Update the version information in plugin.xml files along with lerna version bumps for package.json files

https://issues.jboss.org/browse/AGCORDOVA-232

## Description
These are the approaches considered:

1. Make all plugin.xml files have a version like `version="VERSION"` or some other replaceable tag. This could interfere with SDK development, but is a bit more reliable on the automation side.
2. Have a regex search for `version="0.0.0..."$` and make sure it is the only version attribute in the xml that is at a line end (there are other `version` attributes in the xml for dependencies). This is more hacky and a little brittle but also more transparent.
3. Have a full-fledged template library generate the final `plugin.xml` and rename the original `plugin.xml.(template)`. This is a little overkill, the simplest template lib I know if is [dot](http://olado.github.io/doT/index.html)

Currently I'm using approach 2, but started off with 1 in 6b1e439


paolobueno:
Also sed's regex flavor is a quite weird!

wtrocki:
>  This could interfere with SDK development, but is a bit more reliable on the automation side.

Idea is great, but let's not merge it until we will make sure that this label works with local dev.

package.json
```diff
@@ -11,9 +11,10 @@
 		"lint": "tslint '*/*/src/**/*.ts' '*/*/test/**/*.ts'",
 		"build": "lerna run build",
 		"watch": "lerna run watch",
-		"prep-release": "./scripts/prepareRelease.sh",
-		"validate-release": "./scripts/validateRelease.sh",
-		"publishCanary": "lerna publish --canary",
+		"release:prep": "./scripts/prepareRelease.sh",
+    "release:validate": "./scripts/validateRelease.sh",
+    "updateXml": "lerna exec -- \\$LERNA_ROOT_PATH/scripts/updatePluginXmlVersion.sh",
```
josemigallas: I don't know if we should have this as a npm script, since it is not meant to be run standalone, or is it?

package.json
```diff
@@ -11,9 +11,10 @@
 		"lint": "tslint '*/*/src/**/*.ts' '*/*/test/**/*.ts'",
 		"build": "lerna run build",
 		"watch": "lerna run watch",
-		"prep-release": "./scripts/prepareRelease.sh",
-		"validate-release": "./scripts/validateRelease.sh",
-		"publishCanary": "lerna publish --canary",
+		"release:prep": "./scripts/prepareRelease.sh",
+    "release:validate": "./scripts/validateRelease.sh",
```
josemigallas: Fix extra space lines 14-15 🔍

josemigallas:
LGTM despite two minor comments. I also would prefer option 1 but it looks problematic because the `plugin.xml` is not transpiled or bundled, so that it would have the `VERSION` replaced, then published, then lastly restore it to `VERSION` again?

wtrocki:
@paolobueno See what script @darahayes done for ios release it actually allows us to have versions without problem of committing them into repo.

package.json
paolobueno: Actually not sure why this is tab-indented lol

package.json
paolobueno: The current approach wouldn't make any changes if ran by itself on updated files, we can remove it from package.json and use lerna exec directly from the build scripts but I personally see no harm

paolobueno:
@wtrocki you mean https://github.com/aerogear/aerogear-ios-sdk/blob/master/scripts/prep-release.sh#L7 ? I can't see what difference it has.

josemigallas:
> @josemigallas yes, we would have to watch out not to commit the updated file after a publish.
> 
> But if we completely automate the release process to be similar to what we do in the metrics service for example, CI would make that change locally before the publish and we wouldn't have that concern.

@paolobueno each release tags the corresponding commit with the exact version being released, hence the `VERSION` will be substituted and then committed anyway, won't it?

PR #85, opened 2018-05-09
Use array of checks instead of varadic param

aidenkeating:
## Motivation

Fixes a weird compilation issue with the return type that for some reason didn't show up in CircleCI.

## Description
Replace varadic parameter with array


aidenkeating:
@wtrocki I had the same experience. Must be something to do with versions or something. Anyway, this seems to work for everyone.

PR #86, opened 2018-05-09
Top level api for JavaScript SDK

wtrocki:
## Motivation

Provide seamless api that will allow to initialize configuration and services only once. Allow to initialize services behind the scenes (as required for push and metrics)

## How
```
import { init } from '@aerogear/app';
let mobileServiceConfig = require("../mobile-services.json");
init(mobileServiceConfig);

// And things will work now :)
```


packages/auth/src/AuthService.ts
```diff
@@ -13,16 +13,14 @@ export class AuthService {
   private auth: KeycloakInstance;
   private internalConfig: any;
 
-  constructor(appConfig: AeroGearConfiguration) {
-    const configuration = new ConfigurationHelper(appConfig).getConfigByType(AuthService.TYPE);
-
+  constructor() {
```
wei-lee: in other SDKs, we have the `MobileCore` instance passed in as part of the constructor. I am wondering if we should do the same. Makes it bit easier to test, and also make it obvious that the service modules do depends on the core modules.

packages/core/src/Core.ts
```diff
@@ -0,0 +1,53 @@
...
+   * @param config configuration that should be injected to all available SDK's
+   */
+  public init(config: AeroGearConfiguration): Promise<void> {
+    if (!config || !config.services || config.services.length === 0) {
+      return Promise.reject("Invalid configuration format");
+    }
+    this.configurations = config.services;
+
+    return new Promise((resolve, reject) => {
+      this.metrics = new MetricsService();
```
wei-lee: I know this is how it's done in other SDKs, but I am wondering given this is javascript and we get event emitters for free, should we not considering let the core emit a "ready" event, and then use the app module to init the metrics module.

It also means that if there are other services want to be inited automatically in the future, we don't have to do it in the core.

packages/auth/src/AuthService.ts
wtrocki: @wei-lee Basing on what was agreed on list there will be work done on Android and Xamarin to not expose core. I intentionally done this without it.

packages/core/src/Core.ts
wtrocki: Yes. We will also need to do the same for Push SDK. SDK init is just initialization of the configuration so events may be little bit too much, however I will experiment with this when initializing Push SDK. Push SDK work will allow us to even more improve this and use events.

wtrocki:
@wei-lee TL;DR - I will apply all your suggestions.

docs/core/core.adoc
```diff
@@ -30,14 +30,11 @@ Refer to the link:./getting-started[Getting Started] guide for more information
 
 == Usage
 
-The Core module supports parsing of the configuration file generated by the mobile CLI, in practice this feature is used internally by other modules of the SDK however:
+The Core module supports parsing of the configuration file generated by the mobile CLI, in practice this feature is used internally by other modules of the SDK:
```
josemigallas: Don't like the `however`, @wtrocki ? :P

docs/core/core.adoc
```diff
@@ -30,14 +30,11 @@ Refer to the link:./getting-started[Getting Started] guide for more information
...
 
 [source, javascript]
 ----
-import { ConfigurationHelper } from '@aerogear/core';
-import appConfiguration from '../mobile-services.json';
-
-const helper = new ConfigurationHelper(appConfiguration);
-const metricsConfig = helper.getConfigByType(MetricsService.TYPE);
+import { coreInstance } from '@aerogear/core';
```
josemigallas: I think we should remove this part, it is not user docs, it is something for internal development.

packages/app/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@aerogear/app",
+  "version": "0.2.1",
+  "description": "",
+  "main": "dist/index.js",
+  "types": "types/index.d.ts",
```
josemigallas: should we better point at the folder instead of file?

docs/core/core.adoc
wtrocki: Very funny, **however** we need to be adults here :D

packages/app/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "homepage": "https://github.com/aerogear/aerogear-js-sdk#readme",
+  "devDependencies": {
+    "mocha": "^5.1.1",
+    "@types/chai": "^4.1.3",
+    "chai": "^4.1.2",
+    "nyc": "^11.7.1"
+  },
+  "dependencies": {
+    "@aerogear/core": "0.2.1"
```
josemigallas: If we have core as dependency, and app will be always installed, could we have core as "peerDependency" in modules? (auth, push, etc)

packages/app/src/index.ts
```diff
@@ -0,0 +1,11 @@
+
+import { AeroGearConfiguration, coreInstance } from "@aerogear/core";
+
+/**
+ * Initialize all platform SDK's by using configuration generated by AeroGear command line client
+ *
+ * @param config object containing SDK. Object can be read from file or pasted directly in source code.
+ */
+export const init = function(config: AeroGearConfiguration): Promise<void> {
```
josemigallas: Could we use fat arrow function here?

docs/core/core.adoc
wtrocki: @josemigallas This will be done separately to get this out of the official docs: https://issues.jboss.org/browse/AEROGEAR-2692

packages/auth/src/AuthService.ts
```diff
@@ -13,16 +13,14 @@ export class AuthService {
...
+    const configuration = coreInstance.getConfigByType(AuthService.TYPE);
     if (!configuration || configuration.length === 0) {
       console.warn("Keycloak configuration is missing. Authentication will not work properly.");
       this.internalConfig = {};
     } else {
       this.internalConfig = configuration[0].config;
     }
-
```
josemigallas: Better looking with this blank line

packages/app/package.json
wtrocki: Comment of the year! I'm going to check that!

packages/auth/test/indexTest.ts
```diff
@@ -0,0 +1,5 @@
+import { assert, expect } from "chai";
```
josemigallas: What's this `indexTest` file? I think it should be `AuthService.ts`.

packages/core/src/index.ts
```diff
@@ -1,11 +1,12 @@
 /**
  * @module @aerogearservices/core
+ *
+ * Module used to aggregate common functionalities for AeroGear Services SDK's
  */
 import console from "loglevel";
 console.setDefaultLevel(console.levels.WARN);
 
 // Configuration parsers
-export * from "./configuration";
```
josemigallas: Why renaming now?

packages/core/src/index.ts
```diff
@@ -1,11 +1,12 @@
...
+export * from "./config";
 
-// Native platform metrics
-export *  from "./metrics";
+export * from "./Core";
```
josemigallas: I think we shouldn't export `AgsCore`, only `coreInstance`.

packages/core/src/metrics/MetricsService.ts
```diff
@@ -19,18 +21,19 @@ export class MetricsService {
 
   protected publisher?: MetricsPublisher;
   protected configuration?: ServiceConfiguration;
-  private readonly defaultMetrics: Metrics[];
-
-  constructor(appConfig: AeroGearConfiguration) {
-    const configuration = new ConfigurationHelper(appConfig).getConfigByType(MetricsService.TYPE);
-    this.defaultMetrics = this.buildDefaultMetrics();
+  private readonly defaultMetrics?: Metrics[];
```
josemigallas: I thought the consensus was to have defaultMetrics no matte what?

packages/core/src/metrics/MetricsService.ts
```diff
@@ -19,18 +21,19 @@ export class MetricsService {
...
     if (configuration && configuration.length > 0) {
+      this.defaultMetrics = this.buildDefaultMetrics();
       this.configuration = configuration[0];
       this.publisher = new NetworkMetricsPublisher(this.configuration.url);
       this.sendInitialAppAndDeviceMetrics();
     } else {
-      console.warn("Metrics configuration is missing. Metrics will not be published to remote server.");
+      console.warn("Metrics configuration is missing." +
+        "Metrics will not be published to remote server.",
+        coreInstance);
```
josemigallas: Not sure about printing coreInstance, it is for internal use. Maybe print just the configs?

packages/core/src/metrics/MetricsService.ts
```diff
@@ -63,7 +66,7 @@ export class MetricsService {
 
     const { publisher } = this;
 
-    if (!publisher) {
+    if (!publisher || !this.defaultMetrics) {
       console.info("Metrics server configuration is missing. Metrics will be disabled.");
       return Promise.resolve();
```
josemigallas: Not for this PR but perhaps we should return that msg in the promise as `err` instead of printing in the console.

packages/core/test/CoreTest.ts
```diff
@@ -0,0 +1,57 @@
+import { assert, expect } from "chai";
+import mocha from "mocha";
+import { AeroGearConfiguration, ServiceConfiguration } from "../src/config";
+import { AgsCore } from "../src/Core";
+import testConfigJson from "./mobile-config.json";
+
+describe("AeroGear Core Tests", () => {
```
josemigallas: Let's use the tested class' name: `AgsCore`.

packages/auth/src/AuthService.ts
wtrocki: That was ts-lint error :)

packages/core/test/CoreTest.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const aerogearConfig = testConfigJson as AeroGearConfiguration;
+  let parser: AgsCore;
```
josemigallas: Not a parser anymore, let's call it what it is: `core`.

packages/core/test/CoreTest.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  beforeEach(() => {
+    parser = new AgsCore();
+    parser.init(aerogearConfig);
+  });
+
+  describe("#constructor", () => {
+
+    it("should throw if configuration is null", () => {
+      parser.init(null).then(() => {
```
josemigallas: It is not catching a throw. To do it:
```
const test = () => parser.init(null);

expect(test).toThrow("...");
```

packages/core/src/metrics/MetricsService.ts
wtrocki: IF metrics are disabled then makes no sense. However we can check if they are enabled before creating object. Then logic will be much simpler and we will avoid optionals.

packages/core/src/metrics/MetricsService.ts
wtrocki: How I did not see that :)

packages/core/src/metrics/MetricsService.ts
wtrocki: I think if we check that before creating metrics then we will avoid having all of this :)

packages/core/test/CoreTest.ts
```diff
@@ -0,0 +1,57 @@
...
+import testConfigJson from "./mobile-config.json";
+
+describe("AeroGear Core Tests", () => {
+
+  const aerogearConfig = testConfigJson as AeroGearConfiguration;
+  let parser: AgsCore;
+
+  beforeEach(() => {
+    parser = new AgsCore();
+    parser.init(aerogearConfig);
```
josemigallas: if init returns promise, let's use `async/await` so that beforeEach is fully executed.

packages/core/test/CoreTest.ts
```diff
@@ -0,0 +1,57 @@
...
+      assert.equal(configurations, services);
+    });
+  });
+
+  describe("#getConfig", () => {
+
+    it("should return undefined if using an nonexistent key", () => {
+      const result = parser.getConfigByType("foo");
+
+      assert.isArray(result);
```
josemigallas: should return `undefined` or empty Array?

wtrocki: Test title is wrong. Empty array :)

packages/core/test/metrics/MetricsServiceTest.ts
```diff
@@ -2,7 +2,8 @@ import { assert, expect } from "chai";
 import mocha from "mocha";
```
josemigallas: This is under `test` folder already, then the file should be named the same as the file it is testing.

packages/core/src/metrics/MetricsService.ts
josemigallas: @wtrocki I thought if metrics was disable (i.e. no config present), metrics would be logged in the console anyway. But don't get me wrong, it is OK for me having this as optional, I was just curious.

packages/core/src/metrics/MetricsService.ts
josemigallas: The problem is `publish` is a public method so that the user can call it, hence we have to have this check.

packages/core/src/metrics/MetricsService.ts
wtrocki: > metrics would be logged in the console anyway.

packages/core/src/metrics/MetricsService.ts
wtrocki: @josemigallas You are right about the method but user will not have access to metrics just other SDK's that use core.

packages/core/src/Core.ts
paolobueno: If we settle on the singleton pattern it might be even better to have the init call idempotent and have the promise resolve immediately when core is ready. This way we avoid corner cases of late subscribers to the event emitter.

Adding small code example in follow up comment

wtrocki:
@paolobueno LOL - I was about to go exactly the same path. We need to have promise because init is triggered on user side. I will need to refactor auth to initialize only when promise is satisfied.

packages/core/test/metrics/MetricsServiceTest.ts
wtrocki: Then it's hard to get if you working with test or actual file. Mockito do not provide valid stacktraces so it's also hard to get where error happened.

packages/app/package.json
wtrocki: Done.

packages/core/src/index.ts
wtrocki: Moving to singleton. Still required

packages/core/src/Core.ts
```diff
@@ -9,21 +9,21 @@ import { MetricsService } from "./metrics";
 export class AgsCore {
 
   public configurations?: ServiceConfiguration[];
-  private metrics?: MetricsService;
+  public metrics?: MetricsService;
```
josemigallas: readonly?

packages/core/src/metrics/MetricsService.ts
```diff
@@ -93,10 +85,20 @@ export class MetricsService {
     });
   }
 
+  /**
+   * Collect metrics for all active metrics collectors
+   * Send data using metrics publisher
+   */
+  protected sendAppAndDeviceMetrics(): Promise<any> {
+    return this.publish(MetricsService.DEFAULT_METRICS_TYPE, []).catch((error) => {
```
josemigallas: Do we want to hide this error?

packages/core/test/CoreTest.ts
```diff
@@ -4,35 +4,41 @@ import { AeroGearConfiguration, ServiceConfiguration } from "../src/config";
...
   const aerogearConfig = testConfigJson as AeroGearConfiguration;
-  let parser: AgsCore;
+  let core: AgsCore;
 
-  beforeEach(() => {
-    parser = new AgsCore();
-    parser.init(aerogearConfig);
+  beforeEach(async () => {
+    core = new AgsCore();
+    core.init(aerogearConfig);
```
josemigallas: needs `await` here.

wtrocki:
I have created separate PR with singleton based api. Auth is singleton etc.

With this aproach auth is always available and we using observer to get notification about config. However because values are initialized after this aproach is causing some additional overhead for checking.

PR here: https://github.com/wtrocki/aerogear-js-sdk/pull/1

@josemigallas I'm worried that this PR may grow to the state where it will not easy to merge it. Can we sync to get most of your comments addressed.

wtrocki:
Ready for final review/verification

packages/app/src/index.ts
```diff
@@ -0,0 +1,11 @@
+
+import { AeroGearConfiguration, coreInstance } from "@aerogear/core";
+
+/**
+ * Initialize all platform SDK's by using configuration generated by AeroGear command line client
+ *
+ * @param config object containing SDK. Object can be read from file or pasted directly in source code.
+ */
+export const init = (config: AeroGearConfiguration): void => {
+  return coreInstance.init(config);
```
paolobueno: `void` seems wrong here, isn't it `Promise<void>`?

packages/auth/src/Auth.ts
```diff
@@ -1,28 +1,27 @@
-import { AeroGearConfiguration, ConfigurationHelper, ServiceConfiguration } from "@aerogear/core";
+import { coreInstance, ServiceConfiguration } from "@aerogear/core";
```
paolobueno: I vote for `INSTANCE` over `coreInstance` :)

packages/app/src/index.ts
wtrocki: This has huge implication for users that need to move to async api. It makes it really cumbersome to work with angular. This api do not require anything that is async as it is just setting up configuration.

wtrocki:
@aidenkeating @StephenCoady @rachael-oregan There is some critical bug currently on cordova based plugins. See 72d5051 . We should apply the same for security stuff.

rachael-oregan:
@wtrocki I'll add this into the security checks, nice catch :)

packages/core/src/metrics/platform/CordovaAppMetrics.ts
```diff
@@ -17,16 +19,29 @@ export class CordovaAppMetrics implements Metrics {
...
-      .then(results => ({
-        appId: results[0],
-        appVersion: results[1],
-        sdkVersion: version
-      }));
+    return new Promise((resolve, reject) => {
+      if (!document) {
+        return Promise.reject("Metrics not running in browser environment");
+      }
+      document.addEventListener("deviceready", () => {
```
paolobueno: maybe refactor the event subscription into something like `platformUtils.whenDeviceReady().then(() => {...})`

PR #89, opened 2018-05-11
mobdoc19 add antora structure

finp:
adding antora structure to docs, 
need to follow up with another PR where I move files around, but want to merge this one just to validate publishing

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,5 +34,8 @@
 		"tslint": "5.8.0",
 		"typedoc": "^0.11.1",
 		"typescript": "^2.7.2"
+	},
+	"dependencies": {
+		"@antora/site-generator-default": "^1.0.1"
 	}
 }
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -34,5 +34,8 @@
 		"tslint": "5.8.0",
 		"typedoc": "^0.11.1",
 		"typescript": "^2.7.2"
+	},
+	"dependencies": {
+		"@antora/site-generator-default": "^1.0.1"
```
wtrocki: dev dependencies

finp:
@wtrocki  that edit to package.json was an error on my part, oops.
You don't need antora to edit an existing doc, but you need @antora/site-generator-default to validate your changes locally as per https://github.com/aerogear/processes/blob/master/modules/docs/pages/antora.adoc if you are changing navigation, creating new files, that type of thing

wtrocki:
> aerogear/processes:modules/docs/pages/antora.adoc@master

Let's post that on list.

PR #93, opened 2018-05-14
Update getting-started docs

josemigallas:
## Description
Some of the docs were outdated and others had outdated and/or duplicated info. With this PR I try to create a coherent and concise getting started guide.

There is a general guide `getting-started` that is the starting point for everyone wanting to start using AeroGear. This file describes basic steps to install and initialize the SDK.

At the bottom of it, a section points to the different Services a user can make use of. In each of those files, only specific and relevant steps are described.

For example, there is no need of explaining every time how to get the configuration file or how to initialize the application.

## Verification
For verification, please avoid reviewing this PR changes directly. Checkout the branch and navigate through them using your IDE or explorer of preference.

Check the following:
* The format among the files is coherent
* Titles and sections structure is tidy
* The information is correct and easy to follow
* There are no duplicated steps


docs/modules/getting-started/pages/metrics.adoc
```diff
@@ -0,0 +1,69 @@
+= AeroGear Services Metrics SDK
+
+This document explains how to use and integrate the AeroGear Metrics SDK on a Cordova application.
+
+=== Install Dependencies
+
+Install the necessary dependencies in the app by running the following commands in your terminal:
+
+[source.bash]
+npm install @aerogear/app @aerogear/core
```
wtrocki: Just app. Core will be required anyway.

docs/modules/getting-started/pages/metrics.adoc
```diff
@@ -0,0 +1,69 @@
...
+cordova plugin add @aerogear/cordova-plugin-aerogear-metrics
```
wtrocki: This is only when you have metrics server so I will not consider this an generic step.

docs/modules/getting-started/pages/metrics.adoc
wtrocki: Missing description what this dependency means.

docs/modules/getting-started/pages/metrics.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+=== Deploy Metrics in Openshift
+
+Deploy an instance of Metrics from the Openshift console.
+
+=== Get the App Configuration
+
+The configuration needed to initialize the AeroGear SDK is contained in a JSON file generated by the https://github.com/aerogear/mobile-cli[AeroGear Mobile CLI]. It has the necessary parameters for every deployed service as well as other global ones. It should look like:
```
wtrocki: Invalid link format. `https://github.com/aerogear/mobile-cli[AeroGear Mobile CLI]`. At least do not render in preview.

docs/modules/getting-started/pages/metrics.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+```
```
wtrocki: IMHO we do not need config example.

docs/modules/getting-started/pages/metrics.adoc
```diff
@@ -0,0 +1,69 @@
...
+      ...
+    },
+    ...
+  ]
+}
+```
+
+=== Usage
+
+After dependencies are installed, call `init` from `@aerogear/app` when the app just starts:
```
wtrocki: I will add that `Init method will configure and initialize all other AeroGear SDK's`

docs/modules/getting-started/pages/metrics.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+```
+import { init } from "@aerogear/app";
+
+const appConfig = { ... };
+
+document.addEventListener("deviceready", () => init(appConfig), false);
```
wtrocki: No longer needed to have deviceready. We can have generic case that will work for web and cordova. Deviceready event was specific for metrics but I hidden that inside SDK to simplify user interactions

docs/modules/getting-started/pages/metrics.adoc
```diff
@@ -0,0 +1,69 @@
+= AeroGear Services Metrics SDK
```
wtrocki: Can we drop metrics in name and title? This should be generic getting started that could also have notes about metrics.

docs/modules/getting-started/pages/metrics.adoc
```diff
@@ -0,0 +1,69 @@
...
+import { init } from "@aerogear/app";
+
+const appConfig = { ... };
+
+document.addEventListener("deviceready", () => init(appConfig), false);
+```
+
+As part of the SDK initialization, the relevant app and device data will be sent to the server automatically. Metrics is designed to send the data only on app startup, hence it shouldn't (and can't) be used manually.
+
+=== Provided Metrics
```
wtrocki: IMHO this duplicates existing metrics documentation in other PR. Best to just have links here to other inidividual SDK's.

docs/modules/getting-started/pages/metrics.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+The default metrics contain app and device metrics and possibly other types will be added in the future.
```
wtrocki: IMHO this duplicates existing metrics documentation in other PR. Best to just have links here to other individual SDK's. We can pick one here as example (probably the custom runtime connector once is available)

paolobueno:
@josemigallas @wtrocki I updated the location for some of the files that were still outside of the antora project structure and addressed @wtrocki 's current comments in https://github.com/josemigallas/aerogear-js-sdk/pull/1, I'll link these as resolution for the relevant JIRA.

docs/modules/getting-started/pages/metrics.adoc
josemigallas: What's the point then, if you don't have metrics, you wouldn't even consider looking at this documentation.

docs/modules/getting-started/pages/metrics.adoc
josemigallas: `link` is meant for relative links whereas this format is for external ones (see [reference](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#links))

docs/modules/getting-started/pages/metrics.adoc
josemigallas: I think a visual aid is always helpful, actually I would consider adding it into the code snippet directly (instead of what's now `const config = { ... }`.

docs/modules/getting-started/pages/metrics.adoc
josemigallas: IMO out of the scope, It's the metrics sdk's getting-started guide. Actually I'm thinking, maybe the best way would write a "getting started with the SDK", and then a small "usage" guide for each individual one. Think that otherwise we have to explain the `init` method in every pacakge (metrics getting-started, auth getting-started, push.. etc).

We could have a main page called "How to get started with SDK" or "Integrate AeroGear in your app" with these steps. This way also, we wouldn't include the plugin installation as required step (as you suggests in a comment above). @paolobueno @wtrocki WDYT?

docs/modules/getting-started/pages/metrics.adoc
wtrocki: Point was to build getting started. We do not need metrics getting started etc.

PR #100, opened 2018-05-16
fix: add script and docs for link

wtrocki:
## Motivation

New version of lerna no longer using npm link. 
This means that to use modules outside (in showcase app ) we need to link them manually. @TommyJ1994 Hit issue when link was pointing to different source code. This PR resolves that.

Pinging various people just for information about new workflow in repo.
@aidenkeating @ciaranRoche @StephenCoady @TommyJ1994 @wei-lee 

## Verification

```
npm run link
```


josemigallas:
👀

PR #101, opened 2018-05-16
Add ClientId Handler / Use Top Level Keycloak URL

TommyJ1994:
## Motivation
Add ClientId Handler / Use Top Level Keycloak URL

## Progress
- [x] Add client ID handler (keycloak.js uses clientId, not resource)
- [x] Use top level keycloak URL for proxy/DMZ use

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -23,6 +23,7 @@ export class Auth {
       this.internalConfig = configuration[0].config;
       // create a resource field containing the clientID. The keycloak JS adapter expects a clientId.
       this.internalConfig.clientId = this.internalConfig.resource;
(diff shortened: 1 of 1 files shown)


packages/auth/src/Auth.ts
```diff
@@ -21,6 +21,9 @@ export class Auth {
       this.internalConfig = {};
     } else {
       this.internalConfig = configuration[0].config;
+      // create a resource field containing the clientID. The keycloak JS adapter expects a clientId.
+      this.internalConfig.clientId = this.internalConfig.resource;
+      this.internalConfig.url = configuration[0].url
```
wtrocki: Let's map that to separate field that is type of ServiceConfiguration.

Issue #106, opened 2018-05-21
Security and Metrics Interdependency


evanshortiss:
## Description
I'm building a sample application that uses the Security functionality and calls `checkManyAndPublishMetric`. I was running into an issue that stated "Missing required plugin to collect metrics".

The issue was that I had not run `npx ionic cordova plugin add @aerogear/cordova-plugin-aerogear-metrics --save` so the `app-version` plugin was missing.

Might be necessary to call this out somehow? It was easy to miss.

## Steps to reproduce
1. Install the following:

```
    "@aerogear/app": "~0.3.0",
    "@aerogear/cordova-plugin-aerogear-security": "~0.3.0",
    "@aerogear/security": "~0.3.0",
```

2. Initialise using standard pattern from showcase app
3. Call `checkManyAndPublishMetric`
4. Error `Missing required plugin to collect metrics` is returned

wtrocki:
We need to add to documentation that metrics plugin is also required for security module:
```
cordova plugin add @aerogear/cordova-plugin-aerogear-metrics
```
Link to https://github.com/aerogear/aerogear-js-sdk/blob/master/docs/modules/getting-started/pages/metrics.adoc will be enough.

wei-lee:
@wtrocki should the metrics plugin be declared as a dependency of the security plugin?

wtrocki:
We have it documented as if you have metrics server in config you need this plugin. If we declare it for security then we may get cordova warnings on cordova prepare. It should be ok to have it as dependency but I think is easier to explain that having metics config requires that plugin. Secuirty checks can run without metrics so it will be optional anyway.

wtrocki:
IMHO we need to add reference to metrics doc from secuirty docs. Adding dependency by default is ok for me.

evanshortiss:
I was thinking similar to @wei-lee about adding it as a dependency, but it's only a dependency **if** you choose to send the metric data so technically it optional, right?

Perhaps adding the documentation as you suggested, but also update the error message from `"Missing required plugin to collect metrics"` to `"Missing required plugin to collect metrics. Verify cordova-plugin-aerogear-security is installed."` would be a good compromise?

wei-lee:
+1 on @evanshortiss's suggestion. We can make the error message a bit more clear.

PR #107, opened 2018-05-23
Improve error output and doc for security metric publishing

evanshortiss:
## Motivation
See issue #106 

## Description
See issue #106 

## Progress
- [ ] Unit tests
- [x] Documentation


evanshortiss:
@wtrocki I marked as WIP due to unit tests but looks like there's no existing test to update so I think this is good to go? Increased testing is probably out of scope here.

PR #109, opened 2018-05-24
security package is depending on old version of core

rachael-oregan:
## Motivation

Security package is depending on old version of core

## Description

Updated the package to depend on core@0.3.0

## Progress
- [x] - version bump
- [ ] - release new version of security package
- [ ] - update all the other packages to the same version?


rachael-oregan:
out of date

wtrocki:
This PR changes peer dependency that should not be used. We use classical dependency that has been updated. Going to close it

PR #110, opened 2018-05-25
UPS PUSH SDK

wtrocki:
## Motivation

Provide initial idea how Cordova Push SDK should look like.

Add base wrapper for Cordova plugin that will be used to stage changes that are not merged/released upstream. Add push registration method that can be used in cordova.

## Tasks

- [x] Registration plugin
- [x] Plugin wrapper


pb82:
Do you also want to wrap the `registerDevice` event in there (allow users to provide a handler for that)? Or does this just expose the plugin itself? In that case, what's the purpose of the wrapper?

wtrocki:
@pb82 This wrapper will be used to lock the version of the plugin and being able to replace it in case of issues with something else. JS package will be used only for registration.

packages/push/src/index.ts
```diff
@@ -0,0 +1,11 @@
+
+import { AeroGearConfiguration, INSTANCE } from "@aerogear/core";
+
+/**
+ * Initialize all platform SDK's by using configuration generated by AeroGear command line client
+ *
+ * @param config object containing SDK. Object can be read from file or pasted directly in source code.
+ */
+export const init = (config: AeroGearConfiguration): void => {
```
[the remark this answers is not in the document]
wtrocki: Yeah. Going to push this changes soon.

wtrocki:
@pb82 This branch should work end to end with the plugin. 
I can push this changes to upstream for better collaboration if neeed.

packages/push-cordova/README.md
```diff
@@ -0,0 +1,19 @@
+## AeroGear Metrics Cordova plugin
```
danielpassos: s/Metrics/Push

packages/push-cordova/README.md
```diff
@@ -0,0 +1,19 @@
...
+Wrapper for existing push plugin phonegap-plugin-push
+
+See push plugin documentation for more details:
+https://github.com/phonegap/phonegap-plugin-push
+
+
+## Installation
+
+```
+cordova plugin add @aerogear/cordova-plugin-aerogear-push
```
danielpassos: Any reason why not use the conversation aerogear-[plataform]-[feature]?

eg: aerogear-cordova-push-plugin

packages/push-cordova/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@aerogear/cordova-plugin-aerogear-push",
+  "version": "0.3.0",
```
danielpassos: 0.3.0?

packages/push/README.md
```diff
@@ -0,0 +1,5 @@
+## AeroGear Services Unified Push Registration SDK
```
danielpassos: I think we should remove _Registration_ because it's not _only_  about _Registration_. Wdyt?

packages/push/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@aerogear/push",
+  "version": "0.3.0",
```
danielpassos: 0.3.0?

packages/push/src/PushRegistration.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  private pushConfig: any;
+
+  constructor() {
+    const configuration = INSTANCE.getConfigByType(PushRegistration.TYPE);
+    if (!configuration || configuration.length === 0) {
+      console.warn("Push configuration is missing. UPS server registration will not work.");
+      this.pushConfig = {};
+    } else {
+      this.pushConfig = configuration[0];
```
danielpassos: I really don't like this option, I see we also have this on other platforms, IMO the developer should provide the ID of the push he/she wanna use.

packages/push-cordova/README.md
wtrocki: If you mean `convention` we use cordova plugin convention for discoverability in npm.

packages/push-cordova/package.json
wtrocki: We use single version per entire SDK. New packages will get the version that currently exist.

packages/push/README.md
wtrocki: It is only about registration, all other functionalities will be handled in the plugin. The whole purpose of our SDK should be to enable/simplyfy UPS registration

packages/push/src/PushRegistration.ts
wtrocki: @danielpassos This behavior was agreed on mailing list. Feel free to post this there. We cannot change this without changing other SDK's etc.

packages/push/src/PushRegistration.ts
```diff
@@ -0,0 +1,81 @@
...
+   * @param deviceToken token that will be sent to Unified Push server
+   * @param alias device alias used for registration
+   * @param categories coma separate list of categories that device should be register to.
+   */
+  public register(deviceToken: string, alias: string = "", categories: string = ""): Promise<void> {
+    if (!window || !window.device) {
+      return Promise.reject("Registration requires cordova plugin. Verify the " +
+        "@aerogear/cordova-plugin-aerogear-metrics plugin is installed.");
+    }
+    if (!this.pushConfig && !this.pushConfig.config && !this.pushConfig.url) {
```
pb82: Should this be `||` instead of `&&`? What if it does have `url` but not `config`? This condition would not fail in such a case.

packages/push/src/PushRegistration.ts
```diff
@@ -0,0 +1,81 @@
...
+      Promise.reject("UPS registration: configuration is invalid");
```
pb82: missing `return`?

packages/push/src/PushRegistration.ts
```diff
@@ -0,0 +1,81 @@
...
+      platformConfig = this.pushConfig.config.ios;
+    } else {
+      return Promise.reject("UPS registration: Platform is not supported.");
+    }
+
+    if (!platformConfig) {
+      return Promise.reject("UPS registration: Platform is configured." +
+        "Please add UPS variant and generate mobile - config.json again");
+    }
+    const authToken = window.btoa(`${platformConfig.variantId}:${platformConfig.variantSecret}`);
```
pb82: Are you sure that the whole string should be encoded and not only the individual parts? Looking at other UPS requests it could be that what you need is more like:

```
const auhtToken = `${window.btoa(platformConfig.variantId)}:${window.btoa(platformConfig.variantSecret)}`;
```

wtrocki:
@pb82 @darahayes This is ready for review. I have tested this changes on IOS platform.

packages/push/src/PushRegistration.ts
wtrocki: I reverted that changes. So this comment appeared again. It works this way.

wtrocki:
@pb82 I need to add unit tests and then going to merge it.

packages/push-cordova/package.json
danielpassos: It means all SDK should have the same version?

wtrocki: We do not care as this version is being changed by our automation scripts on release. Then we decide what versions all packages should have.

pb82:
@wtrocki Ok, can we also publish this version to npm?

wtrocki:
@pb82 @darahayes Added tests. Added additional check if device token is empty. Fixed bug for promises (wrapped into error). IMHO this is really solid now. Can you do quick take look.

wtrocki:
@pb82 I think we can release it.

PR #112, opened 2018-05-30
fix: return promise directly from register call

darahayes:


wtrocki:
This could be good change but it requires us to return non void type from method.

wtrocki:
Closing as outdated. Feel free to create new one.

Issue #105, opened 2018-05-21
Root Detection for Android


evanshortiss:
## Description
I've tested the root checks on my Nexus 5 device and found that our default check is not sufficient for certain devices.

We are invoking `IRoot.isRooted()`, but on my device this doesn't catch the root. I need to invoke `IRoot. isRootedRedBeerWithoutBusyBoxCheck()` which correctly detected the root as did `IRoot.isRootedRedBeer()`.

I'm not 100% sure what solution is best here from an Android perspective. Maybe check both `isRootedRedBeer` and `isRootedRedBeerWithoutBusyBoxCheck`? If the `isRootedRedBeerWithoutBusyBoxCheck` returns false then assume it's a false positive per [this doc](https://github.com/scottyab/rootbeer/blob/master/README.md#false-positives)?

## Steps to reproduce
1. Root or obtain rooted Android Device
2. Invoke `return this.securityService.check(SecurityCheckType.notRooted)`
3. Verify that `check.passed` is false

aidenkeating:
@evanshortiss Just to confirm (all the falses and trues get very confusing 😄).

This check (notRooted) should return:
* `false` if the device is rooted
* `true` if the device is not rooted.

So are you seeing it return `true` when the device is rooted?

I've seen this (https://github.com/scottyab/rootbeer/blob/master/README.md#false-positives) and we should probably take false positives into account. Just confirming the issue you're seeing and the documented issue are the same.

evanshortiss:
@aidenkeating agreed, it is a little confusing! Here's my code:

```ts
  public async ionViewDidEnter () {
    const notRootedCheck = await this.sec.check(SecurityCheckType.notRooted)

    if (notRootedCheck.passed === false) {
      const alert = this.alertCtrl.create({
        title: 'Insecure Device',
        subTitle: 'We detected that this device is rooted. Running as root increases the likelihood of your device being compromised by malicious software designed to steal passwords and sensitive information. Continued use is at your own risk.',
        buttons: ['OK']
      })

      alert.present()
    }
  }
```

If the check fails i.e `notRooted` returns `passed === false` I want to show a warning.

[...]

PR #114, opened 2018-05-31
Fix rooted check false positive

ciaranRoche:
## Motivation

Issue https://github.com/aerogear/aerogear-js-sdk/issues/105

Currently, RootBeer sometimes gives a false positive result to the check. This is often because the manufacturer of the device rom has left the busy box binary.

## Description
We have added a check to the Rootcheck to detect the device platform, as this is a specific problem for android. If the device is an android device we invoke the `isRootedRedBeer` function, this reduces the likely hood of a false positive result. If the device is not an android device we invoke the `isRooted` function.

## Additional Notes
**Important** I have verified this on an Android emulator only, this will need to be verified on an actual rooted device as well as ios to ensure it works.

## Verify
Navigate to the device trust page in the showcase app and observe the result, if the device is rooted the check should fail with the message 'device rooted detected' whereas if the device is not rooted the check should pass with the message 'device rooted not detected'

### DIFF
--- a/packages/security/src/deviceTrust/checks/NonRootedCheck.ts
+++ b/packages/security/src/deviceTrust/checks/NonRootedCheck.ts
@@ -1,6 +1,6 @@
+import { isCordovaAndroid } from "@aerogear/core";
 import { SecurityCheck } from "../SecurityCheck";
 import { SecurityCheckResult } from "../SecurityCheckResult";
-import { isCordovaAndroid } from "@aerogear/core";
 
 declare var IRoot: any;
 declare var document: any;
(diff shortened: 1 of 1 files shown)


packages/security/src/deviceTrust/checks/NonRootedCheck.ts
```diff
@@ -28,12 +30,12 @@ export class NonRootedCheck implements SecurityCheck {
...
       document.addEventListener("deviceready", () => {
-        if (!IRoot) {
-          reject(new Error("Could not find plugin IRoot."));
+        if (!IRoot || !device) {
+          reject(new Error("Could not find plugin for Root Check"));
           return;
         }
-
-        IRoot.isRooted((rooted: number) => {
+        const isRootedCheck = device.platform === "Android" ? IRoot.isRootedRedBeer : IRoot.isRooted;
```
wtrocki: Check `isCordovaAndroid` method in core that could be used here.

ciaranRoche: Good catch :+1:

PR #121, opened 2018-06-06
push docs initial draft

pb82:
## Motivation

Documentation for the Cordova Push SDK

## Description
Documentation for installing, configuring and using the Cordova push SDK.


docs/modules/ROOT/pages/push.adoc
```diff
@@ -0,0 +1,94 @@
+== AeroGear JavaScript Unifiedpush SDK
```
wtrocki: `UnifiedPushServer`

docs/modules/ROOT/pages/push.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+An SDK that allows you to receive push notifications and integrate with the Aerogear Unifiedpush Server. In contrast to the other SDKs, `@aerogear/push` itself does not implement the handling of push notifications. It helps you to integrate a third party push library with our backend, the Aerogear Unifiedpush Server. While you can use any push plugin you like that supports Firebase Cloud Messaging and Apple Push Notification Services we suggest link:https://github.com/phonegap/phonegap-plugin-push[Phonegap Plugin Push] for Cordova and Ionic Apps.
```
wtrocki: `In contrast to the other SDKs, `@aerogear/push` itself does not implement the handling of push notifications.` - Other SDK's do not do that either.

wtrocki: `While you can use any push plugin you like that supports Firebase Cloud Messaging and Apple Push Notification Services we suggest link:https://github.com/phonegap/phonegap-plugin-push[Phonegap Plugin Push] for Cordova and Ionic Apps.`

pb82: @wtrocki Do you think i should remove that sentence? Should we even mention the possibility of using other push plugins?

docs/modules/ROOT/pages/push.adoc
wtrocki: Some of the instructions bellow are strictly Cordova related. IMHO it's best to just call that Cordova SDK. If not we should provide documentation for push registration only and possibly have separate page for plugin. WDYT?

docs/modules/ROOT/pages/push.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+=== Installation
+
+For a Cordova App two packages need to be installed: `@aerogear/push` and the platform push plugin:
+
+[source,bash]
+----
+$ npm install --save @aerogear/push
+$ cordova plugin add phonegap-plugin-push
```
wtrocki: That is not right. We should install our plugin

docs/modules/ROOT/pages/push.adoc
pb82: You mean, it would be possible to use @aerogear/push in a react-native app, but then you would obviously need to use a different push implementation? Yes i guess we can separate those the docs with regards to this.

docs/modules/ROOT/pages/push.adoc
```diff
@@ -0,0 +1,94 @@
...
+      "config": {
+        "ios":      <iOS specific config>
+        "android":  <Android specific config>
+      }
+    }
+  ]
+}
+----
+
+Next you need to configure the push provider itself. `phonegap-plugin-push` defaults to Firebase Cloud Messaging on Android and iOS. Login to your link:https://firebase.google.com/[Firebase account] and retrieve the config file (`google-services.json` or `google-services.plist`). Include it in the root of your project folder. Please refer to the link:https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/INSTALLATION.md#android-details[Plugin documentation] for more details.
```
wtrocki: Firebase IOS will not work with Unified Push Server as Firebase will hide complexity of the APNS itself and hide crucial elements Push Server needs. Firebase acts as proxy here and sending messages from UPS will not be possible. Let's not mention that.

docs/modules/ROOT/pages/push.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+=== Usage
+
+To make use of the Aerogear Unifiedpush Server, devices need to be registered with it. This in handled by `@aerogear/push` and should be performed after the platform push plugin itself has finished initialization:
```
wtrocki: `To make use of the Aerogear Unifiedpush Server, devices need to be registered with it` => Unified Push Server allows sending messages to all registered devices, independent of the platform that is used.

docs/modules/ROOT/pages/push.adoc
pb82: Ok, whats the name of the config file that you need to include for iOS?

docs/modules/ROOT/pages/push.adoc
wtrocki: Plus let's add. Please refer to plugin documentation for more info.

docs/modules/ROOT/pages/push.adoc
```diff
@@ -0,0 +1,94 @@
...
+    // Registration with UPS successful, you can now send push notifications from the UPS UI
+  }).catch(err => {
+    // Error on device registration
+  });
+});
+----
+
+==== Options
+
+You can specify a number of options for the registration. The signature of `PushRegistration#register` is:
```
wtrocki: -1. Intelisense, api docs can be used here. This can outdated etc.

wtrocki: Actually let's leave it. You put hard work to get that done so no sense to remove it.

pb82: I don't mind removing the code snippet, just wanted to explain the parameters. But maybe they should be documented in the UPS docs instead?

docs/modules/ROOT/pages/push.adoc
wtrocki: That's tough one. We can introduce here to the topic and create Cordova chapter. We do not need separate file. Then if React Native will be there we can just add another chapter.

docs/modules/ROOT/pages/push.adoc
wtrocki: > Should we even mention the possibility of using other push plugins?

No. We always tend to mention what we test and support only.

docs/modules/ROOT/pages/push.adoc
wtrocki: IOS do not require any config file.

docs/modules/ROOT/pages/push.adoc
wtrocki: We have something here: https://github.com/aerogear/aerogear-js-sdk/blob/master/packages/push/src/PushRegistration.ts#L34-L40

docs/modules/ROOT/pages/push.adoc
darahayes: One thing that really sticks out like a sore thumb is the sentence `It helps you to integrate a third party push library with our backend, the Aerogear Unifiedpush Server.` The use of the word `backend` here is especially misleading.

To be honest I find this entire summary to be quite confusing. I would be more straight to the point and I would try to break it up. It's a pretty large paragraph.

Something like this:


== AeroGear JavaScript Unifiedpush SDK

`@aerogear/push` makes it simple to receive push notifications in Cordova and Ionic Apps. It is a thin wrapper that extends the popular link:https://github.com/phonegap/phonegap-plugin-push[Phonegap Plugin Push] library to also work with Aerogear Unifiedpush Server.

It can be integrated with Aerogear Unifiedpush Server or it can be used directly with Firebase Cloud Messaging and Apple Push Notification Service.

=== Supported Platforms

* Android
* iOS

pb82: Yep, thats better. Going to use this.

docs/modules/ROOT/pages/push.adoc
```diff
@@ -0,0 +1,95 @@
...
+      "config": {
+        "ios":      <iOS specific config>
+        "android":  <Android specific config>
+      }
+    }
+  ]
+}
+----
+
+Next you need to configure the push provider itself. For Android, login to your link:https://firebase.google.com/[Firebase account], download the config file (`google-services.json`) and include it in the root of your project folder. On iOS it should work out of the box.
```
wtrocki: On IOS you need to configure application to use push ID. I will not mention anything about IOS here as we point people to plugin documentation!

PR #124, opened 2018-06-07
Documentation improvements based on feedback

wtrocki:
## Motivation

Improving documentation based on product owner input

### DIFF
--- a/docs/modules/ROOT/pages/push.adoc
+++ b/docs/modules/ROOT/pages/push.adoc
@@ -85,7 +85,7 @@ pushObject.on('registration').subscribe(data => {
 });
 
 // Notification handler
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/push.adoc
```diff
@@ -82,6 +83,11 @@ pushObject.on('registration').subscribe(data => {
     // Error on device registration
   });
 });
+
+// Notification handler
+PushService.pushObject.on('notification').subscribe(notification => {
```
pb82: should be: `pushObject.on` (without `PushService`, that is showcase specific).

wtrocki: Awesome catch! We may point people to showcase application on review. This will reduce number of issues in docs code snippets.

PR #126, opened 2018-06-08
ci: allow any tag to kick off release process

darahayes:
CircleCI does a bad job of validating these tags. It is also redundant here because we do our own validation as part of the release. Much better for the release to kick off and fail over the release not ever kicking off because that would leave people very confused


wtrocki:
Fully agree! Really good change! Random tag is not going to cause release as it will not match the x.x.x format!

PR #128, opened 2018-06-11
Enable Custom Runtime Connector in app

wtrocki:
## Motivation

Minor change to enable custom runtime connector queries in app.


wtrocki:
@pb82 FYI

packages/app/src/index.ts
```diff
@@ -9,3 +9,12 @@ import { AeroGearConfiguration, INSTANCE } from "@aerogear/core";
 export const init = (config: AeroGearConfiguration): void => {
   return INSTANCE.init(config);
 };
+
+/**
+ * Get a custom runtime connector configuration
+ *
+ * @param id - id of the custom runtime connector
+ */
+export const getCustomRuntimeConfigById = (id: string): ServiceConfiguration | undefined => {
```
david-martin: Does this match what's being done in other sdks to allow fetching custom runtime config?

wtrocki: Not sure what you mean? Nothing was done in other SDK's to support it.

david-martin: I'm not sure. Is this a 'syntactic sugar' function (which is fine), or is it a necessary function for a dev to even get access to the crc config?
If former, maybe other sdks could have it too.
If latter, then it sounds like we need it and cordova is a special case.

Either way, I don't mind.
As long as we're thinking about other SDKs and if they need to/should change too.

wtrocki: This is only syntactic sugar function, but it's looks better in documentation.

PR #129, opened 2018-06-12
Fix is rooted check by invoking android check function

aidenkeating:
## Motivation
Currently we are not invoking the Android check function in the
rooted check. The object we're checking will always evaluate to
true, meaning we're always trying to perform an Android-specific
check. Causing the check to crash on iOS.

## Description
This fixes the check by invoking the function.

## Progress
- [x] Invoke the Android check function


aidenkeating:
@ciaranRoche @StephenCoady Mind taking a look? This should fix the rooted check issue in https://github.com/aerogear/cordova-showcase-template/pull/55

ciaranRoche:
:eyes:

StephenCoady:
Still seeing 3/4 checks on iOS

screenshot 2018-06-12 13 22 47

PR #131, opened 2018-06-13
Add required variables to phonegap plugin

wtrocki:
## Motivation

Made changes to push plugin in order to get things installed.


wtrocki:
@danielpassos This is required to release push!

wtrocki:
TL;DR - parameters are breaking things and specifying them in the dependency do not make sense - we will lock them to something that users do not want to have. Users can override parameters now. Change can be verified by using `1.0.0-dev-2`. 

ping @psturc

psturc:
Verified with `version="1.0.0-dev-3`.

PR #132, opened 2018-06-22
Add check to getConfigByType function

ciaranRoche:
## Motivation

Currently, if there is a blank service object in the `mobile-service.json` it causes the SDK to fall over,  by handling it and returning the object as false, the core remains stable and the missing config is picked up by our individual modules on init.

## Verify
- Remove one of the service objects and replace with a blank object in the example app
- Run the app, and it should crash on startup
- Symlink this PR changes to the example app.
- Rerun the app, it will not crash on startup

### DIFF
--- a/packages/core/src/Core.ts
+++ b/packages/core/src/Core.ts
@@ -30,7 +30,12 @@ export class AgsCore {
    */
   public getConfigByType(type: string): ServiceConfiguration[] | undefined {
     if (this.configurations) {
-      return this.configurations.filter(service => service.type.toLowerCase() === type.toLowerCase());
+      return this.configurations.filter(service => {
+        if (!service.type) {
+          return false;
+        }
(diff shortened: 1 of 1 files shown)


packages/core/src/Core.ts
```diff
@@ -41,7 +41,7 @@ export class AgsCore {
    */
   public getConfigById(id: string): ServiceConfiguration | undefined {
     if (this.configurations) {
-      return find(this.configurations, service => service.id.toLowerCase() === id.toLowerCase());
+      return find(this.configurations, service => !!service.type && service.id.toLowerCase() === id.toLowerCase());
```
wei-lee: why we are checking `service.type` here? Should it be `service.id && service.id.toLowerCase()`

ciaranRoche: good catch @wei-lee copy and paste oversight :+1:

PR #134, opened 2018-06-27
Release version 1.0.0-alpha.1

ciaranRoche:
## Motivation

We need to push a release to test some small changes and fixes ahead of the community release

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -8,5 +8,5 @@
       "exact": true
     }
   },
-  "version": "1.0.0-alpha"
+  "version": "1.0.0-alpha.1"
 }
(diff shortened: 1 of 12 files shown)


darahayes:
Totally cool with doing a release, although I'm not so sure about the version number. I would probably do 1.0.1-alpha but I don't know what the others would think about this considering we haven't done our overall community release yet. Going to post a question on the mailing list.

aidenkeating:
@darahayes This follows the prerelease spec in https://semver.org/#spec-item-9 and the precedence spec in https://semver.org/#spec-item-11. We follow the same approach for Android and iOS.

PR #136, opened 2018-06-28
WIP: Aerogear 3296

StephenCoady:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-3296

This PR fixes an issue where (in a development environment) it is currently not possible to login from a cordova app as the cordova adapter [here](https://github.com/keycloak/keycloak-js-bower/blob/master/dist/keycloak.js#L1289) always uses localhost.

**The Solution**
This PR, along with https://github.com/aerogear/cordova-showcase-template/pull/78 solves this problem by:

1. Forcing the keycloak js adapter to use the "default" adapter.
2. Makes some changes to keycloak by making a function used by init publicly available
3. Handling the flow of login/logout slightly differently (due to the fact we now lose some context upon switching between the browser and the app)

The major drawback of this solution is that it requires us to have our own fork of the Keycloak JS adapter. This is because since we are using it in a way they don't prescribe the chances of us getting any changes merged upstream are slim.

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -1,7 +1,20 @@
-import { INSTANCE, ServiceConfiguration } from "@aerogear/core";
-import Keycloak from "keycloak-js";
-import { KeycloakInitOptions, KeycloakInstance, KeycloakProfile } from "keycloak-js";
-import console from "loglevel";
+// @ts-ignore
+import {
+  INSTANCE,
+  ServiceConfiguration
+} from "@aerogear/core";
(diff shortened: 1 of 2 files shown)


packages/auth/src/Auth.ts
```diff
@@ -14,6 +27,12 @@ export class Auth {
   private auth: KeycloakInstance;
   private internalConfig: any;
 
+  private initOptions: any;
```
josemigallas: So many `any`s.... 😆

PR #138, opened 2018-07-04
Refactor docs script

ciaranRoche:
## Motivation
Currently, the docs script falls over due to the Cordova plugins not containing an `src` directory. 
## Description
This change clones the docs.aerogear.org repo into the docs directory,  it pipes a different config to the `lerna.json`, this specifies the modules we want to generate documentation for. It runs the typedoc script and places the output in the appropriate location within the docs.aerogear.org repo. Before completing it pipes back in the old config to the `lerna.json` file
## Verify
- Checkout PR
- Run `npm run generate:docs`
- Navigate to docs -> docs,aerogear.org -> api -> cordova -> latest to view the new documentation
- Rerun `npm run generate:docs`
- The script should run with no issue.

### DIFF
--- a/scripts/docs.sh
+++ b/scripts/docs.sh
@@ -16,15 +16,8 @@ else
     rm -rf $DOCS_LOCATION
 fi
 
-echo "Piping Config to lerna.json"
-# Ensure the package names are correct
-echo "{\"lerna\": \"2.9.0\",\"packages\": [\"packages/app\",\"packages/auth\",\"packages/core\",\"packages/push\",\"packages/security\"],\"commands\": {\"publish\": {\"exact\": true}},\"version\": \"1.0.0-alpha.1\"}" > ${PWD}/lerna.json 
-
 echo "Generate documentation for packages"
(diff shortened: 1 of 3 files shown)


scripts/docs.sh
```diff
@@ -2,13 +2,29 @@
...
+    echo "Moving repo to docs folder"
+    mv ${PWD}/$DOC_FOLDER ${PWD}/docs/
+else 
+    echo "Removing existing documentation"
+    rm -rf $DOCS_LOCATION
+fi
+
+echo "Piping Config to lerna.json"
+# Ensure the package names are correct
+echo "{\"lerna\": \"2.9.0\",\"packages\": [\"packages/app\",\"packages/auth\",\"packages/core\",\"packages/push\",\"packages/security\"],\"commands\": {\"publish\": {\"exact\": true}},\"version\": \"1.0.0-alpha.1\"}" > ${PWD}/lerna.json
```
wtrocki: you can use filter in lerna command line to exclude packages that we do not want to have (they all have cordova in name so it's easy)

wtrocki: https://github.com/lerna/lerna#--scope-glob

wtrocki:
It works but it will be easier to use scope.

ciaranRoche:
@wtrocki nice catch, used the `--ignore` tag, mind having another look when you get a chance

PR #144, opened 2018-08-24
Fixed updatePluginXmlVersion.sh to make it works on MAC

ziccardi:
## Motivation

The script didn't work on mac because `sed -i` on mac requires an additional parameter, not needed on GNU `sed`


wtrocki:
Good pick. Tried and tested on old bash.

@ziccardi by default mac is using ancient version of bash shell. Ideally if you use shell on daily basis it will be recommended to use something more modern like oh-my-zsh :)

PR #149, opened 2018-10-08
Update github files

camilamacedo86:


.github/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
wtrocki: Missing REDHAT ?

camilamacedo86: Let's keep in this way for know since all already update and it is more a detail. We can edit this file in the standards and run again.

.github/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
+
```
wtrocki: Let's remove license. We do not want to change project license without notifying users etc.
All aerogear projects have license files in the ROOT

camilamacedo86: @wtrocki are not all with the LICENSE and it should be equals for all projects in this way was agreed to keep in this dir. Is easier for us if we centralized all in one place. However, please, feel free to remove it from the standards then we re-run the script.

PR #150, opened 2018-11-05
Sync sdk

wtrocki:
## Motivation

Feature branch with all required functionalities for sync sdk. 
We will use this to stage work for Sync SDK.


wtrocki:
Wrong branch

PR #151, opened 2018-11-05
Sync sdk feature branch

wtrocki:
# Sync Feature branch

Sync SDK will using Apollo Link extensible chain to build set of processors for various SDK operations.
This processors will be automatically configured to be build and give users access to sync client object.

screen shot 2018-11-06 at 1 07 42 pm

## Note

Let's use this branch as base for PR's

## Verification

https://github.com/aerogear/offline-conflict-strategies/pull/34


packages/sync/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/aerogear/aerogear-js-sdk.git"
+  },
+  "keywords": [
+    "aerogear",
+    "aerogear services",
+    "graphql"
+  ],
+  "author": "Aero Gear",
```
danielpassos: AeroGear (without space)

packages/sync/src/DataSyncClientOptions.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface DataSyncClientOptions {
+  /**
+   * Describes name of the field used as ID
+   */
+  defaultObjectIdField?: string;
+
+  /**
+   * Storage solution
+   */
+  storage?: PersistentStore<PersistedData>;
```
danielpassos: Love it!

wtrocki: If you like this like this PR upstream: https://github.com/helfer/apollo-link-queue
Hopefully we could get this merged and get official support for persistence.

packages/sync/src/DataSyncClientOptions.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Contains all options required to initialize SDK
+ * @see DefaultOptions for defaults
+ */
+export interface DataSyncClientOptions {
+  /**
+   * Describes name of the field used as ID
+   */
+  defaultObjectIdField?: string;
```
danielpassos: I think we should not use `default` in this field.

packages/sync/src/OptionalUiEngine.ts
```diff
@@ -0,0 +1,7 @@
+
+/**
+ * Engine for handling optional UI
+ */
+export class OptionalUIEngine {
```
danielpassos: Not clear what is this for.

PR #152, opened 2018-11-06
Core config

wtrocki:
## Motivation
 
Add base for core config.


packages/sync/src/ClientCreator.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  await persistCache({
+    cache,
+    storage
+  });
+  return apolloClient;
+};
+
+
+validate
```
StephenCoady: missing function here

wtrocki: Yeah. Was in the rush forgot to save before commit.

PR #155, opened 2018-11-12
added base for conflict link

StephenCoady:
NOTE: Stalling work on this PR to focus on tickets being brought into the spring.

## Motivation
https://issues.jboss.org/browse/AEROGEAR-8091

## What
Added a base for conflict detection. Also exports two basic strategies for use by the client.

## Verification

First clone this branch and from within `packages/sync` run `npm link .`

Then from within the offline strategies repo, on this branch: https://github.com/aerogear/offline-conflict-strategies/pull/34

1. Run the server:
```
cd server
docker-compose up -d
npm i
npm run start
```

2. Run the app:
```
cd web
npm i
npm link @aerogear/datasync-js
npm run start 
```

3. Trigger the conflict:
When the app is running open two separate browser windows (must not both be the same browser - or one should be incognito).
Create a user and have it displayed on both windows.
Edit the user on one window.
Try to edit the user on the second window. The second window should win and the conflict should be displayed in the consoles.

## Progress

- [x] Add basic conflict detection using Apollo link error
- [x] Add two basic strategies
- [ ] Add flag to check if the server has resolved the conflict before it returns to the client
- [ ] Add a notification handler to always return information about the conflict to the client, possibly through some hook to the client
- [ ] Implement the conflict link in the createClient file so that it can be used/tested with a client

### DIFF
--- a/packages/sync/src/conflicts/conflictLink.ts
+++ b/packages/sync/src/conflicts/conflictLink.ts
@@ -20,7 +20,7 @@ export const conflictLink = (config: DataSyncConfig) => {
                 }
             }
         }
-    }
+    };
     /**
      * Fetch the conflict strategy if one is provided, if not return client wins.
      */
@@ -30,7 +30,7 @@ export const conflictLink = (config: DataSyncConfig) => {
         } else {
             return strategies.diffMergeClientWins;
         }
-    }
+    };
 
     return onError(({ operation, forward, graphQLErrors }) => {
         const data = getConflictData(graphQLErrors);
--- a/packages/sync/src/conflicts/conflictLink.ts
+++ b/packages/sync/src/conflicts/conflictLink.ts
@@ -1,3 +1,4 @@
+import { ApolloLink, FetchResult, NextLink, Observable, Operation } from "apollo-link";
 import { onError } from "apollo-link-error";
 import { GraphQLError } from "graphql";
 import { DataSyncConfig } from "../config/DataSyncConfig";
@@ -19,8 +20,7 @@ export const conflictLink = (config: DataSyncConfig) => {
                 }
             }
         }
-    };
-
+    }
     /**
      * Fetch the conflict strategy if one is provided, if not return client wins.
      */
@@ -30,9 +30,9 @@ export const conflictLink = (config: DataSyncConfig) => {
         } else {
             return strategies.diffMergeClientWins;
         }
-    };
+    }
 
-    return onError(({ graphQLErrors, operation, forward }) => {
+    return onError(({ operation, forward, graphQLErrors }) => {
         const data = getConflictData(graphQLErrors);
         const resolvedConflict = getConflictStrategy()(data, operation.variables);
         // TODO Notify
@@ -40,5 +40,4 @@ export const conflictLink = (config: DataSyncConfig) => {
         operation.variables = resolvedConflict;
         return forward(operation);
     });
-
 };
--- a/packages/sync/src/conflicts/conflictLink.ts
+++ b/packages/sync/src/conflicts/conflictLink.ts
@@ -1,0 +1,51 @@
+import { ApolloLink } from "apollo-link";
+import { onError } from "apollo-link-error";
+import { GraphQLError } from "graphql";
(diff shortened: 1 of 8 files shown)


wtrocki:
@StephenCoady IMHO PR seems to be complete as it offers:
Ability to detect and handle conflict. Provide different handlers.
IMHO we can merge that and just create tasks for the rest of the unfinished items.

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -0,0 +1,51 @@
+import { ApolloLink } from "apollo-link";
+import { onError } from "apollo-link-error";
+import { GraphQLError } from "graphql";
+import { DataSyncConfig } from "../config/DataSyncConfig";
+import { ConflictResolutionData, strategies } from "./strategies";
+
+export class ConflictLink {
```
wtrocki: [Optional] Any reason why this cannot be link itself (extend Apollo Link)?

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -0,0 +1,51 @@
...
+    constructor(config: DataSyncConfig) {
+        this.config = config;
+    }
+
+    /**
+     * Initialise the conflict link
+     * @returns An apollo link capable of detecting errors
+     */
+    public init(): ApolloLink {
+        const link = onError(({ graphQLErrors, operation, forward }) => {
```
wtrocki: Antipattern.  Init is not operating on class and not initializing anything in properties.

wtrocki: If we extend the link we can include any other link inside. For example:
https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-ws/src/webSocketLink.ts#L49-L53

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -0,0 +1,51 @@
...
+            const data = this.getConflictData(graphQLErrors);
+            if (!this.config.conflictStrategy) {
```
wtrocki: I will extract that if to separate method that returns conflictStrategy function

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -0,0 +1,51 @@
...
+                this.config.conflictStrategy = strategies.diffMergeClientWins;
+            }
+            const resolvedConflict = this.config.conflictStrategy(data, operation.variables);
```
wei-lee: is this function always a synchronised one? Is it possible to support async ones?

wtrocki:
Very minor comments. I will verify that after changes.

packages/sync/src/conflicts/conflictLink.ts
StephenCoady: @wei-lee It looks like onError doesn't support async functions at the moment. https://github.com/apollographql/apollo-link/issues/646#issuecomment-401399248

wei-lee: ok, does this mean it will not be possible to allow developers to implement a strategy where it may involve building some UIs to let user choose?

StephenCoady: @wei-lee it looks like Apollo relies on Observables to do things like this. I am trying to get it working with these at the moment so I will update once I know for sure.

wtrocki: onError do not need to be async. We could process everything in observable/promise that will deliver resolved object and pass it forward into chain.  

On the other note we might also need to support different conflictResolver per type or at least enable devs to handle that case

packages/sync/src/conflicts/conflictLink.ts
StephenCoady: @wtrocki I spent yesterday evening trying to get this work as a class. I implemented it using observables under the request method but for some reason I could not get it to work as expected. When I tried to do it manually I could not trigger the errors to run the logic we need and when I tried to wrap onError in the request I lost access to graphQLErrors which we need to execute the same logic. I am reverting this to a function for now so we can merge and possibly come back to it.

PR #157, opened 2018-11-19
Add NetworkStatus

danielpassos:
Add a very simple wrapper for online/offline listeners.

PS: I have tested it in so many browsers and devices.


danielpassos:
Just wondering if it should be part of sync or core and if we should make it more Apollo Link like

danielpassos:
@StephenCoady I have tried to make the `fn` a `Function` but `window.addEventListener` expect a `[boolean, any, String]`

StephenCoady:
@danielpassos do you think we could point this PR to my branch instead? that way when it is all ready to go to master (with tests) it will all go at the same time? it just doesn't make much sense to merge this on its own.

danielpassos:
@StephenCoady Sure. Go ahead!

packages/sync/src/NetworkStatus.ts
```diff
@@ -0,0 +1,25 @@
...
+   *
+   * @param fn Function to be called when got offline
+   */
+  public static whenOffline(fn: any): void {
+    window.addEventListener("offline", fn, false);
+  }
+
+}
+
+export default NetworkStatus;
```
wtrocki: Avoid default exports. 
This is not going to work with javascript and it will require `const interface = require("x").networkInterface.default`

packages/sync/src/NetworkStatus.ts
```diff
@@ -0,0 +1,25 @@
+class NetworkStatus {
```
wtrocki: This class is pointless. It has only static fields.

wtrocki:
I believe that we need interface for Network status with cordova and web satisfying this interface. 
This PR is not giving us that.

danielpassos:
The intention was to create a separated class which is responsible to provide a when to call something when got online and offline to unblock @StephenCoady on https://github.com/aerogear/aerogear-js-sdk/pull/158

So...

1. Should I replace the `NetworkStatus` into an interface with those 2 methods that the frameworks/platforms will implement?
2. Should I move the implementations to somewhere?

wtrocki:
Practically we need to support cordova/web (plus have plan for react native). 
To do that effectively. We need to crete typescript interface and supply implementations for platform.
If you looking for examples see how we are doing metrics - cordova implementation is extracted and specific to the cordova etc. To help here I will create interface and supply cordova implementation for this. This is really simple task  - max 10 minutes of work so no worries.

PR #158, opened 2018-11-19
Mutation Queue functionality

StephenCoady:
Adding this to separate PR on top of @wtrocki just in case it is no good and needs to be changed.

This is a top level branch which will hold all functionality for offline queue mutations.

The work to test theseSDK changes is currently being progressed on this [PR](https://github.com/aerogear/offline-conflict-strategies/pull/35) of the sample application.

### Tasks:
- [x] Build Queue mechanism
- [x] Connect the storage to this offline queue 
- [x] Persist this queue after the client is restarted
- [x] Add squashing interface to allow us to squash common requests (opt out feature, not opt in)
- [x] Add isSquashable directive to provide schema option to use squashing
- [x] Add onlineOnly directive to provide a schema option to only allow certain operations when the client is online
- [ ] Add notification handler for online only operations
- [ ] Add support for optimistic responses when offline
- [ ] Add support for delete mutations when offline
- [ ] Documentation for usage
- [ ] Integration tests

### Verification:
TODO

### Issues: 

Apollo link provide a debounce package [here](https://github.com/helfer/apollo-link-debounce)
However it is not currently maintained (and hasn't been for a year) and so does not work with the current Apollo link we are using (1.2.3) and instead uses (0.7.0). This means any debounce links we create with this package will not work.

As a workaround I have taken the debounce code and imported it directly into our package. There is currently a [PR](https://github.com/helfer/apollo-link-debounce/pull/4) upstream which aims to fix this but it has been open for months and I am unsure why. I will leave this custom debouncer where it is until we know more about upstream.


packages/sync/src/createClient.ts
```diff
@@ -38,12 +46,12 @@ function extractConfig(userConfig: DataSyncConfig | undefined) {
  */
 async function buildStorage(clientConfig: DataSyncConfig) {
   const cache = new InMemoryCache({
-    dataIdFromObject: () =>  clientConfig.dataIdFromObject
+    dataIdFromObject: (object) =>  object.id
```
wtrocki:
```suggestion
    dataIdFromObject: (object) =>  {
      if( clientConfig.dataIdFromObject) {
          return clientConfig.dataIdFromObject(object)
      } 
      return object.id
     }
```

packages/sync/src/createClient.ts
```diff
@@ -14,10 +15,17 @@ import { PersistedData, PersistentStore } from "./PersistentStore";
...
-
-  return new ApolloClient({ link , cache });
+  const { cache, storage } = await buildStorage(clientConfig);
+  const link = buildLink(clientConfig, storage);
+  const options: ApolloClientOptions<NormalizedCacheObject> = {
+    link,
+    cache
+  };
+  const apolloClient = new ApolloClient<NormalizedCacheObject>(options);
+  const syncOfflineMutations = new SyncOfflineMutation(apolloClient, storage, "offline-mutation-store");
```
wtrocki: "offline-mutation-store" - let's use config for this with default one pointing to this value.

packages/sync/src/directives/onlineOnly.ts
```diff
@@ -0,0 +1,15 @@
+import { SchemaDirectiveVisitor } from "graphql-tools";
```
wtrocki: Invalid file name/to class `OnlineOnlyDirective`

packages/sync/src/links/DebounceLink.ts
```diff
@@ -0,0 +1,176 @@
+import {
```
wtrocki: If this is used for squashing then -1. Debouncing implementation has inside it's own queue which will complicate testing and development of this functionality. We are kinda replicating the code. Original code for debouncing implemented in spike will be more specialized and we can integrate that carefully as part of the link-queue interface.

packages/sync/src/links/LinksBuilder.ts
```diff
@@ -1,35 +1,44 @@
...
+  (config: DataSyncConfig, storage: PersistentStore<PersistedData>): ApolloLink => {
     if (config.customLinkBuilder) {
-      return config.customLinkBuilder(config);
+      return config.customLinkBuilder(config, storage);
     }
 
     const httpLink = new HttpLink({ uri: config.httpUrl });
     // TODO drop your links here
     let compositeLink;
+    const queueMutationsLink = new QueueLink(storage, "offline-mutation-store");
```
wtrocki: Externalize this string from config

packages/sync/src/links/QueueLink.ts
```diff
@@ -32,10 +41,8 @@ export default class QueueLink extends ApolloLink {
     }
 
     public request(operation: Operation, forward: NextLink ) {
-        if (this.isOpen) {
-            return forward(operation);
-        }
-        if (operation.getContext().skipQueue) {
+        // TODO split this conditional and add a handler to notify of online only cases
+        if (this.isOpen || hasDirectives(["onlineOnly"], operation.query)) {
```
wtrocki: constant this directive (no need for config - we need constants file to have some central piece where we put stuff that is indirectly connected to server.

packages/sync/src/createClient.ts
```diff
@@ -14,10 +15,17 @@ import { PersistedData, PersistentStore } from "./PersistentStore";
...
+  await syncOfflineMutations.init();
```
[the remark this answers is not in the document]
StephenCoady: @wtrocki this needs to stay because init needs to use await.

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -5,39 +5,41 @@ import { DataSyncConfig } from "../config/DataSyncConfig";
...
+  /**
+  * Fetch conflict data from the errors returned from the server
+  * @param graphQLErrors array of errors to retrieve conflicted data from
+  */
+  const getConflictData = (graphQLErrors?: ReadonlyArray<GraphQLError>): ConflictResolutionData => {
+    if (graphQLErrors) {
+      for (const err of graphQLErrors) {
+        if (err.extensions) {
+          // TODO need to add flag to check if conflict was resolved on the server
+          if (err.extensions.exception.type === "AgSync:DataConflict") {
```
wtrocki: We can constant this string

packages/sync/src/createClient.ts
```diff
@@ -1,23 +1,31 @@
...
+  const { cache, storage } = await buildStorage(clientConfig);
+  const link = buildLink(clientConfig, storage);
+  const options: ApolloClientOptions<NormalizedCacheObject> = {
+    link,
+    cache
+  };
+  const apolloClient = new ApolloClient<NormalizedCacheObject>(options);
+  const syncOfflineMutations = new SyncOfflineMutation(apolloClient, storage, clientConfig.mutationsQueueName);
+  await syncOfflineMutations.init();
+  await syncOfflineMutations.sync();
```
wtrocki: I know that I have done that initially. But locking rendering until entire network offline queue will be process can seriously slow down application loading.  Users may see black screen  for a while. We can do that asynchronously in the background

packages/sync/src/offline/SyncOfflineMutation.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public sync = async () => {
+        // if there is no offline data  then just exit
+        if (!this.hasOfflineData()) { return; }
+
+        // return as promise, but in the end clear the storage
+        const uncommittedOfflineMutation: OperationQueueEntry[] = [];
+
+        await Promise.all(this.offlineData.map(async (item) => {
```
wtrocki: It's critical to preserve order of operations. If edit mutations happen before create mutation then some requests will crash.

packages/sync/src/offline/SyncOfflineMutation.ts
```diff
@@ -0,0 +1,75 @@
...
+            try {
+                await this.apolloClient.mutate({
+                    variables: item.operation.variables,
+                    mutation: item.operation.query,
+                    context: item.operation.getContext
+                });
+            } catch (e) {
+                // set the errored mutation to the stash
+                uncommittedOfflineMutation.push(item);
```
wtrocki: Really important line. Mutate here can result error like conflict.  This can be popped back depending on error strategy. Note to myself to test this case if we do not get orphan offline mutations that were actually applied in conflict resolution part.

wtrocki:
@StephenCoady Awesome work! Most of the comments were addressed, however I would like to write more tests and try this out completely. Going to merge it into single branch that we are going to use for collaboration!

wtrocki:
I'm going to reflect unfinished bullet points into jira tickets

PR #160, opened 2018-11-23
Web and Cordova Networks Status

danielpassos:


packages/sync/src/offline/CordovaNetworkStatus.ts
```diff
@@ -0,0 +1,11 @@
+import {NetworkStatus} from "./NetworkStatus";
+
+export class CordovaNetworkStatus implements NetworkStatus {
+  public whenOnline(fn: any): void {
+    document.addEventListener("online", fn, false);
```
wtrocki: Lack of detection for global `document`

packages/sync/src/offline/CordovaNetworkStatus.ts
```diff
@@ -0,0 +1,11 @@
+import {NetworkStatus} from "./NetworkStatus";
+
+export class CordovaNetworkStatus implements NetworkStatus {
+  public whenOnline(fn: any): void {
```
wtrocki: Define callback function and use it everywhere. It could be any.

`public whenOnline(callback: OnlineCallback): `

IMHO `whenOffline` is missleading. How about registerOnlineHandler?
Alternatively we may need to have separate method `isOffline` or something like this to allow direct access to state. Not all components needs to listen.

packages/sync/src/offline/WebNetworkStatus.ts
```diff
@@ -0,0 +1,11 @@
+import {NetworkStatus} from "./NetworkStatus";
+
+export class WebNetworkStatus implements NetworkStatus {
+  public whenOnline(fn: any): void {
+    window.addEventListener("online", fn, false);
```
wtrocki: `window` is undeclared global

packages/sync/src/offline/CordovaNetworkStatus.ts
wtrocki: Actually "me thinking" when online/offline is kinda legacy inferace of the browser. 
Maybe we can just have single callback with the network state information like 
`
{
state: "online" | "offline",
metadata?: {
  type: "WIFI etc." 
}

packages/sync/src/links/LinksBuilder.ts
```diff
@@ -31,9 +31,10 @@ export const defaultLinkBuilder: LinkChainBuilder =
     const debounceLink = new DebounceLink();
     let links: ApolloLink[] = [queueMutationsLink, debounceLink, conflictLink(config), httpLink];
 
-    // TODO to be replaced by network interface check when ready
-    NetworkStatus.whenOnline(() => queueMutationsLink.open());
-    NetworkStatus.whenOffline(() => queueMutationsLink.close());
+    const network = new WebNetworkStatus();
```
wtrocki: We kinda need singleton for this for network status to be accessible in entire SDK (and also outside module

packages/sync/src/config/DefaultConfig.ts
```diff
@@ -1,6 +1,10 @@
 import { INSTANCE, ServiceConfiguration } from "@aerogear/core";
 import { ConfigError } from "./ConfigError";
 import { DataSyncConfig } from "./DataSyncConfig";
+import {NetworkStatus} from "../offline/NetworkStatus";
```
wtrocki: if you use visual studio code or webstorm please consider using editorconfig plugin to support formatting.

wtrocki:
It is kinda nice to see this coming in good shape. Good work!

PR #165, opened 2018-11-27
Fix online status on network status callback

danielpassos:


packages/sync/src/offline/WebNetworkStatus.ts
```diff
@@ -8,12 +8,9 @@ declare var window: any;
  */
 export class WebNetworkStatus implements NetworkStatus {
   public onStatusChangeListener(callback: NetworkStatusChangeCallback): void {
-    const networkInfo = {
-      online: !this.isOffline()
```
wtrocki: So this means that `this.isOffline()` is not working properly. Can we create field to hold this value depending on the  state.  Does event listener is going to get triggered after registration  for setting initial state?

danielpassos: Actually `this.isOffline()` works perfectly fine. The problem is in that case, the method is called when the callback/listener is set not when the callback is called so the state never changes.

wtrocki: Ok. Does event listener is going to get triggered after registration for setting initial state? 
We need that to set queue as open/closed.

danielpassos: Nopz. It will be called **only** when network states change. For this, we have the `networkStatus.isOffline()`

PR #166, opened 2018-11-27
Refactor for offline queue

wtrocki:
## Motivation
 
Move network and storage inside link to make it self contain.
@StephenCoady we need see how we can get offlineQueue restored.

### DIFF
--- a/packages/sync/src/offline/OfflineRestoreHandler.ts
+++ b/packages/sync/src/offline/OfflineRestoreHandler.ts
@@ -22,8 +22,12 @@ export class OfflineRestoreHandler {
     this.storage = storage;
     this.storageKey = storageKey;
   }
-
-  public replyOfflineMutations = async () => {
+  /**
+   * Replay mutations to client.
+   * This operation will help to rebuild Apollo Link observer chain
+   * after page refresh/app restart
+   */
+  public replayOfflineMutations = async () => {
     const stored = await this.getOfflineData();
     if (stored) {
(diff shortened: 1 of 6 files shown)


wtrocki:
@StephenCoady I have seen your PR, but it is actually putting some internal SDK problem on the shoulders of users (by adding new elements to api) so -1 from me. I have tested this aproach and it  looks like this will be all we need for the moment. It may happen that we going to need to refactor that later to cater Optimistic UI changes but let 's not go deep now.   It's really hard to enclose  architecture for link as there is circular dependency ( link requires client and client requires link to be created). I have spiked option to inject client later but it proven to be hacky anway.  I think that this is what we can do for now. It will be cool  to add extra logging to this. I have created extra ticket to cater this work: https://issues.jboss.org/browse/AEROGEAR-8150

packages/sync/src/offline/OfflineRestoreHandler.ts
```diff
@@ -1,22 +1,29 @@
...
-              storage: PersistentStore<PersistedData>, storageKey: string) {
+              storage: PersistentStore<PersistedData>,
+              storageKey: string) {
     this.apolloClient = apolloClient;
     this.storage = storage;
     this.storageKey = storageKey;
   }
 
-  public sync = async () => {
+  public replyOfflineMutations = async () => {
```
StephenCoady: this should be called replayOfflineMutations

PR #167, opened 2018-11-27
moved the sync after restart to be part of queue

StephenCoady:
ping @wtrocki this was what I was thinking originally. let me know what you think.

not sure what the client bootstrapping itself from its own state looks like, so this might not be a runner.


packages/sync/src/createClient.ts
```diff
@@ -3,27 +3,22 @@ import { persistCache } from "apollo-cache-persist";
...
-import { SyncOfflineMutation } from "./offline/SyncOfflineMutation";
 
 /**
  * Factory for creating Apollo Client
  *
  * @param userConfig options object used to build client
  */
-export const createClient = async (userConfig?: DataSyncConfig) => {
-
+export const createClient = async (userConfig?: DataSyncConfig, oldClient?: ApolloClient<NormalizedCacheObject>) => {
```
wtrocki: what is `oldClient` for? It is kinda strange to tell people to pass client to function that creates client.

PR #168, opened 2018-11-28
Queue filtering

wtrocki:
## Motivation
 
Add filtering capabilities. Introduce unit tests. Little bit refactor to allow queue to be open closed by other factor that network. I have added little bit of documentation. We need to put more documentation into code as it's pretty raw at the moment.  

*Still working on unit tests*


StephenCoady:
👀

PR #171, opened 2018-11-29
Prevent the OfflineRestoreHandler to fire a mutation if still offline

danielpassos:
screenshot 2018-11-29 10 29 07

### DIFF
--- a/packages/sync/src/offline/OfflineRestoreHandler.ts
+++ b/packages/sync/src/offline/OfflineRestoreHandler.ts
@@ -2,6 +2,7 @@ import ApolloClient from "apollo-client";
 import { NormalizedCacheObject } from "apollo-cache-inmemory";
 import { PersistentStore, PersistedData } from "../PersistentStore";
 import { OperationQueueEntry } from "../links/OfflineQueueLink";
+import {NetworkStatus} from "./NetworkStatus";
 
 /**
  * Class used to restore offline queue after page/application restarts.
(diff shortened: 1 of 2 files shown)


StephenCoady:
@danielpassos having some trouble verifying this, hitting the following problem:

https://www.dropbox.com/s/l91viq1gv7llnf0/Screenshot%202018-11-29%2016.57.51.png?dl=0

danielpassos:
@StephenCoady That is weird, I give it a try again and it works for me ¯\_(ツ)_/¯

wtrocki:
Eghm... Unit tests ?

wtrocki:
@danielpassos I believe that this is already done. Mutations are executed in the row and if they do not succeed they will still be left in the queue. This operation is not to actually fire and succeed but to actually build observer chain. Without that none of the changes will be replicated when we become online again.

packages/sync/src/offline/OfflineRestoreHandler.ts
```diff
@@ -2,6 +2,7 @@ import ApolloClient from "apollo-client";
 import { NormalizedCacheObject } from "apollo-cache-inmemory";
 import { PersistentStore, PersistedData } from "../PersistentStore";
 import { OperationQueueEntry } from "../links/OfflineQueueLink";
+import {NetworkStatus} from "./NetworkStatus";
 
 /**
  * Class used to restore offline queue after page/application restarts.
```
wtrocki: See comment here. This is not actually build to succeed it is to restore the links. So if we are offline actual queue link will be closed and it will hold off those requests till will become online.

wtrocki:
Actually from screenshot I see that there is in fact bug. What should happen is that once we enqueue things to queue it should detect that things are offline and hold of requests (adding things back to queue),
but when we start app it's not detecting that properly and let request to pass. As result we see failed requests. As we support cordova I think it's critical for us to get some simple Ionic use case done as fast as possible (going to do that once we nail most of the tickets) as this will give us fully functional cordova env for tests.

danielpassos:
@wtrocki The point here is if:

1. I'm offline
2. Fire some mutation which is going to be added to the localstore
3. Close the app/website
4. Open the site again
5. OfflineRestoreHandler will try to fire the mutation (even if you still offline)
6. It will get a network error
7. The mutation is going to be added to the localstore again

My only point with this is trying to prevent the `OfflineRestoreHandler` to try step 5 if the app be open and the device still offline

wtrocki:
Problem is more complex that I initially anticipated. 
We need to have reliable environment for testing that: ionic
We need to have extensive logging to see if queue is closing

danielpassos:
So... we realized that `OfflineRestoreHandler` should never really fire the mutation when the app starts offline, it should be handled/intercepted by `OfflineQueueLink` since the app is offline, but for some reason (a bug) the `OfflineRestoreHandler` fire this _before_ OfflineQueueLink` get in place.

PR #172, opened 2018-11-29
added tests and debug profile

StephenCoady:


package.json
```diff
@@ -39,6 +39,7 @@
     "typescript": "^2.7.2"
   },
   "dependencies": {
-    "@antora/site-generator-default": "^1.0.1"
+    "@antora/site-generator-default": "^1.0.1",
```
wtrocki: Antora is no longer needed. Let's remove it

packages/sync/test/SquashOperation.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+describe("SquashOperations", () => {
+  let opQueue: OperationQueueEntry[] = [];
+  beforeEach(() => {
+    opQueue = [];
+  });
+
+  it("check queue contains single operation", () => {
+    const queue = squashOperations(queueEntry, opQueue);
+    expect(queue.length).eqls(1);
```
wtrocki: [Very optional ] I guess we can also check if things are merged using object assign but that is an stretch goal.

PR #174, opened 2018-11-30
Aerogear 8146

StephenCoady:
Adding noSquash feature as per: https://issues.jboss.org/browse/AEROGEAR-8146

NOTE: The ticket specifies an isSquashable directive as opposed to noSquash which is what I have implemented. I feel this makes more sense as the majority of mutations will be squashable so it is opt out of squashing as opposed to opt in.

To try this out add @noSquash to your query and ensure `directive @noSquash on FIELD` is present in your typeDefs.

If you want, those changes are already on the branch [here](https://github.com/aerogear/offline-conflict-strategies/tree/directives-test)

### DIFF
--- a/packages/sync/src/links/OfflineQueueLink.ts
+++ b/packages/sync/src/links/OfflineQueueLink.ts
@@ -93,7 +93,11 @@ export class OfflineQueueLink extends ApolloLink {
   }
 
   private enqueue(entry: OperationQueueEntry) {
-    this.opQueue = squashOperations(entry, this.opQueue);
+    if (hasDirectives([Directives.NO_SQUASH], entry.operation.query)) {
+      this.opQueue.push(entry);
+    } else {
+      this.opQueue = squashOperations(entry, this.opQueue);
+    }
     this.storage.setItem(this.key, JSON.stringify(this.opQueue));
   }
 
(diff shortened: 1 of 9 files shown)


StephenCoady:
@wtrocki @danielpassos any idea what is wrong with the build? Doesn't look like its related to my changes and I can build locally no problem. 

It is complaining about:

```
The inferred type of \'conflictLink\' cannot be named without a reference to \'apollo-link-http-common/node_modules/apollo-link\'. This is likely not portable. A type annotation is necessary.
```

wtrocki:
@StephenCoady We need to create issue for the server (no assigned to any epic) that this no squash Directive should be added.  We need to have that capacity.

packages/sync/src/links/OfflineQueueLink.ts
```diff
@@ -93,7 +93,11 @@ export class OfflineQueueLink extends ApolloLink {
   }
 
   private enqueue(entry: OperationQueueEntry) {
-    this.opQueue = squashOperations(entry, this.opQueue);
+    if (hasDirectives([Directives.NO_SQUASH], entry.operation.query)) {
```
wtrocki: Ignore me here, but if we adding feature for squashing and we have separate file then this change should probably land there. It's good pattern but let's not change it

packages/sync/test/SquashOperation.test.ts
```diff
@@ -143,4 +94,25 @@ describe("SquashOperations", () => {
     expect(opQueue.length).eqls(3);
 
   });
+
+  it("test noSquash Directive", () => {
```
wtrocki: This is not how unit tests are build. They always tend to test single file. 
here we mixing 2 files in single test.  Actually this directive should be in squashing logic.  queue do not need to be polluted with the  business logic that is not queue oriented.

PR #177, opened 2018-12-03
add onlineOnly tests

StephenCoady:
added some tests for the already existing @onlineOnly directive.

### DIFF
--- a/packages/sync/test/OfflineQueueLink.test.ts
+++ b/packages/sync/test/OfflineQueueLink.test.ts
@@ -8,7 +8,7 @@ import { OfflineQueueLink as QueueLink } from "../src/links/OfflineQueueLink";
 import {
   TestLink
 } from "./TestUtils";
-
+import { opWithOnlineDirective } from "./operations";
 import { expect } from "chai";
 import { PersistentStore, PersistedData } from "../src/PersistentStore";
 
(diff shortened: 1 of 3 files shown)


packages/sync/test/OfflineQueueLink.test.ts
```diff
@@ -98,6 +98,48 @@ describe("OnOffLink", () => {
     expect(testLink.operations.length).eqls(1);
   });
 
+  it("test online only directive skips queue", () => {
+    let operations: any[] = [];
```
wtrocki: [ProTip] Normally this could land in beforeEach method as it common for all tests

packages/sync/test/OfflineQueueLink.test.ts
```diff
@@ -98,6 +98,48 @@ describe("OnOffLink", () => {
...
+        operations = [];
+      },
+      setItem(key: string, content: any) {
+        operations = JSON.parse(content);
+      }
+    };
+    const localConfig = { mutationsQueueName: "test", storage: storageEngine };
+    const queueLink = new QueueLink(localConfig);
+    queueLink.close();
+    const customLink = ApolloLink.from([queueLink, testLink]);
```
wtrocki: [ProTip] We can use test link to see if this was forwarded. If not then we do not need test link at all.

PR #178, opened 2018-12-04
SDK Documentation

StephenCoady:
Added some base documentation to get us started. I tried to cover the top level aspects developers might care about.

Also added a flag to allow the developer to globally disable squashing.


packages/sync/README.md
```diff
@@ -17,3 +17,140 @@ To enable debug please execute in console
...
+} from '@aerogear/datasync-js';
+```
+
+# Configuration
+To provide custom configuration to the client, the following options are available. If you wish, these are also available by using the `DataSyncConfig` interface from the SDK.
+
+```javascript
+declare var window: any;
+
+let config: DataSyncConfig = {
```
wtrocki: Kinda pointless to have those without explanation. So we could just point people to DataSyncConfig file

packages/sync/README.md
```diff
@@ -17,3 +17,140 @@ To enable debug please execute in console
...
+  mutationsQueueName: "offline-mutations-queue",
+  squashing: true
+}
+```
+
+Config | Description | Default
+------ | ------ | ------
+httpUrl | The URL of your http server | N/A
+wsUrl | The URL of your websocket | N/A
+dataIdFromObject | The name of the field to be used as ID | "id"
```
wtrocki: This is not longer used. I will create ticket to restore that to not forget about it

packages/sync/README.md
```diff
@@ -17,3 +17,140 @@ To enable debug please execute in console
...
+
+    let links: ApolloLink[] = [customLink, httpLink];
+
+    let compositeLink = ApolloLink.from(links);
+    return compositeLink;
+  };
+```
+
+## Implementing Custom Network Status checks
+To use your own custom netork checks, implement the NetworkStatus interface which provides two functions;
```
wtrocki: We could add references to github?

packages/sync/README.md
```diff
@@ -17,3 +17,140 @@ To enable debug please execute in console
...
+
+## Offline Mutations
+
+AeroGear Sync SDK provides offline mutations out of the box. By default it uses localStorage to do this and mutations are stored under the cache key of 'offline-mutations-queue'.
+
+### Online Only Queries
+To ensure certain queries are not queued and are always delivered to the network layer, you must make use of Graphql directives. To do so define a directive on your schema type definitions like so:
+
+```
+directive @onlineOnly on FIELD
```
wtrocki: This is no longer problem . This should be defined on server - server should use graphql lower than 14. that allows that. 
I guess we could create ticket on the server and adjust docs accordingly.

WDYT?

packages/sync/README.md
```diff
@@ -17,3 +17,140 @@ To enable debug please execute in console
...
+Next, on your client, ensure the query has the annotation attached like so:
+
+```
+exampleQuery(...) @onlineOnly {
+  ...
+}
+```
+
+### Squashing Queries
+By default AeroGear Sync SDK provides squashing of mutations out of the box. This means that if your client is offline and the same mutation is run twice with different parameters they will be squashed into one mutation. This is beneficial as the client will not have to queue a large amount of mutations to replay once it returns online.
```
wtrocki: Squashing is kinda internal language. I'm reading this as newbie now and first sentence is kinda cryptic. 
We could say:

Changes that are performed on the same object id and mutation name will be joined together when offline. 
Process is called squashing or something like this

packages/sync/src/config/DataSyncConfig.ts
```diff
@@ -48,4 +48,9 @@ export interface DataSyncConfig {
    * The name of the queue to store offline mutations in
    */
   mutationsQueueName?: string | any;
+
+  /**
+   * Whether or not to enable squashing of queries
+   */
+  squashing?: boolean;
```
wtrocki: How about `mergeOfflineMutations` or something less cryptic

packages/sync/README.md
```diff
@@ -17,3 +17,140 @@ To enable debug please execute in console
...
+## Conflict strategies
+You can provide your custom conflict resolution strategies to the client in the config by using the provided `ConflictResolutionStrategy` type from the SDK. An example custom strategy is shown below.
+
+```javascript
+let clientWins = (serverData, clientData) => {
+  return Object.assign(server, client);
+};
+```
+
+To use this strategy pass this function as conflictStrategy in your config object like so:
```
danielpassos: Should we also say we provide some strategies out of the box and how to use this?

packages/sync/README.md
```diff
@@ -17,3 +17,140 @@ To enable debug please execute in console
...
+
+Next, on your client, ensure the query has the annotation attached like so:
+
+```
+exampleQuery(...) @onlineOnly {
+  ...
+}
+```
+
+### Squashing Queries
```
danielpassos: Maybe replace **Queries** by **Mutations** to avoid confusion?

packages/sync/README.md
wtrocki: Not now. We should document only offline stuff for the moment.

packages/sync/README.md
StephenCoady: So lower than version 14 and you don't need to explicitly call it out? Should we not be using the most recent version? I have no preference either way.

PR #180, opened 2018-12-06
added directive link to overall links

StephenCoady:
### Verification

To check this works 
* Run the example application using master of the SDK.
* Do not add any directives to the schema on the server.
* Annotate one of the queries with either @onlineOnly or @noSquash e.g.
`updateUser(dateOfBirth: $dateOfBirth, id: $id, name: $name, version: $version) @onlineOnly`
* In the application run the query you previously annotated.
* You should see an error in the console stating that @onlineOnly or @noSquash is an unknown directive.
* Check out this branch on the SDK
* Leave the example application as before with whichever annotation you chose.
* Run the same query as before
* The query should succeed this time, or at least the directive should not cause an error

### DIFF
--- a/packages/sync/src/links/DirectiveLink.ts
+++ b/packages/sync/src/links/DirectiveLink.ts
@@ -1,0 +1,53 @@
+import {
+  ApolloLink,
+  NextLink,
+  Observable,
+  Operation
+} from "apollo-link";
+import { DirectiveNode } from "graphql";
+import { hasDirectives, removeDirectivesFromDocument, checkDocument } from "apollo-utilities";
+import { Directives, MUTATION_QUEUE_LOGGER } from "../config/Constants";
+import debug from "debug";
+
+export const logger = debug(MUTATION_QUEUE_LOGGER);
+
+export class DirectiveLink extends ApolloLink {
+  constructor() {
+    super();
+  }
+  public request(operation: Operation, forward: NextLink) {
+    logger("IN DIRECTIVE LINK, OPERATION:", operation);
+    debugger;
+    const clientDirectivesPresent = hasDirectives([Directives.ONLINE_ONLY, Directives.NO_SQUASH], operation.query);
+    if (!clientDirectivesPresent) {
+      return forward(operation);
+    } else {
+      const connectionRemoveOnlineOnly = {
+        name: "onlineOnly",
+        test: (directive: DirectiveNode) => directive.name.value === Directives.ONLINE_ONLY,
+        remove: true
+      };
+      const connectionRemoveNoSquash = {
+        name: "noSquash",
+        test: (directive: DirectiveNode) => directive.name.value === Directives.NO_SQUASH,
+        remove: true
+      };
+      checkDocument(operation.query);
+      const newDoc = removeDirectivesFromDocument(
+        [
+          connectionRemoveOnlineOnly,
+          connectionRemoveNoSquash
+        ],
+        operation.query
+      );
+      logger("NEW DOC", newDoc);
+      if (newDoc) {
+        operation.query = newDoc;
+        return forward(operation);
+      }
+      return new Observable(observer => {
+        return () => ({ removedDirective: true });
+      });
+    }
+  }
+}
--- a/packages/sync/src/links/LocalDirectiveFilterLink.ts
+++ b/packages/sync/src/links/LocalDirectiveFilterLink.ts
@@ -4,33 +4,33 @@ import {
   Observable,
   Operation
 } from "apollo-link";
-import { DirectiveNode } from "graphql";
 import { hasDirectives, removeDirectivesFromDocument, checkDocument } from "apollo-utilities";
-import { Directives, MUTATION_QUEUE_LOGGER } from "../config/Constants";
+import { LocalDirectives, MUTATION_QUEUE_LOGGER } from "../config/Constants";
 import debug from "debug";
 
 export const logger = debug(MUTATION_QUEUE_LOGGER);
 
-export class DirectiveLink extends ApolloLink {
+export class LocalDirectiveFilterLink extends ApolloLink {
   constructor() {
     super();
   }
   public request(operation: Operation, forward: NextLink) {
-    logger("IN DIRECTIVE LINK, OPERATION:", operation);
-    debugger;
-    const clientDirectivesPresent = hasDirectives([Directives.ONLINE_ONLY, Directives.NO_SQUASH], operation.query);
+    logger("Checking if client directives need to be removed on ", operation);
+    const clientDirectivesPresent = hasDirectives(
+      [
+        LocalDirectives.ONLINE_ONLY,
+        LocalDirectives.NO_SQUASH
+      ],
+      operation.query
+    );
     if (!clientDirectivesPresent) {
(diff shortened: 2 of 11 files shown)


wtrocki:
I checked code locally and played with this a little. This stuff ended up overly complex for the value we want to get from it.   Initial idea was really simple - strip those out of the document - this could happen in actual offline link where this directives are used. 

Benefits of server side directive initialization:
- Single place to perform queries
- No confusion over directives
- Support for validating queries/playground

StephenCoady:
@wtrocki I am happy to go that direction. I might just outline my thoughts on this approach so they are plain to see.

Pros:
- developer does not have to touch their schema
- they just need to "mark" the query on the client as @onlineOnly or whatever
- the sdk looks after everything and strips this client side directive out before it touches the server

Cons:
- no tracking of these directives on the server
- duplication can occur if they try to implement these directives their queries may behave funny
- possibly more complex (although this complexity has benefits for the end user)

wtrocki:
@StephenCoady Let's get this to work. This convinced me. We may need to document that this directives are not going to work in playground.

packages/sync/src/links/DirectiveLink.ts
```diff
@@ -0,0 +1,53 @@
...
+  Operation
+} from "apollo-link";
+import { DirectiveNode } from "graphql";
+import { hasDirectives, removeDirectivesFromDocument, checkDocument } from "apollo-utilities";
+import { Directives, MUTATION_QUEUE_LOGGER } from "../config/Constants";
+import debug from "debug";
+
+export const logger = debug(MUTATION_QUEUE_LOGGER);
+
+export class DirectiveLink extends ApolloLink {
```
wtrocki: LocalDirectiveFilterLink ?

packages/sync/src/links/DirectiveLink.ts
```diff
@@ -0,0 +1,53 @@
+import {
+  ApolloLink,
+  NextLink,
+  Observable,
+  Operation
+} from "apollo-link";
+import { DirectiveNode } from "graphql";
+import { hasDirectives, removeDirectivesFromDocument, checkDocument } from "apollo-utilities";
+import { Directives, MUTATION_QUEUE_LOGGER } from "../config/Constants";
```
wtrocki: LocalDirectives? to differentiate between server and client?

StephenCoady:
@wtrocki I have fixed it now and its working. You mentioned exporting a single link as one but that might be messy since queueLink needs config. Or is there an easier way I am missing. For now I just have it in the linkBuilder but it doesn't add much to concat in this way.

I know I am missing tests but wanted to push this for now. I need to check that the directives have been removed after the queue link and if they have that should cover us for testing.

StephenCoady:
updated verification steps. will write tests in the morning.

packages/sync/src/links/LocalDirectiveFilterLink.ts
```diff
@@ -0,0 +1,53 @@
...
+} from "apollo-link";
+import { hasDirectives, removeDirectivesFromDocument, checkDocument } from "apollo-utilities";
+import { LocalDirectives, MUTATION_QUEUE_LOGGER } from "../config/Constants";
+import debug from "debug";
+
+export const logger = debug(MUTATION_QUEUE_LOGGER);
+
+export class LocalDirectiveFilterLink extends ApolloLink {
+  private connectionRemoveOnlineOnly = {
+    name: "onlineOnly"
```
wtrocki: LocalDirectives.ONLINE_ONLY,?

packages/sync/src/links/LocalDirectiveFilterLink.ts
```diff
@@ -0,0 +1,53 @@
...
+        LocalDirectives.ONLINE_ONLY,
+        LocalDirectives.NO_SQUASH
+      ],
+      operation.query
+    );
+    if (!clientDirectivesPresent) {
+      return forward(operation);
+    } else {
+
+      checkDocument(operation.query);
```
wtrocki: What this is for? I removed that and it still works as expected.

packages/sync/src/links/LocalDirectiveFilterLink.ts
```diff
@@ -0,0 +1,53 @@
...
+          this.connectionRemoveOnlineOnly,
+          this.connectionRemoveNoSquash
+        ],
+        operation.query
+      );
+      if (newDoc) {
+        operation.query = newDoc;
+        return forward(operation);
+      }
+      return new Observable(observer => {
```
wtrocki: [Question] Why we return Observable in this case? I might not undestand that properly.

If newDoc is null then this means code issue and we should probably throw ugly error to not let that happen. 
Basically newDoc should always be non null - if it is null then there is something seriously wrong.

packages/sync/src/links/LocalDirectiveFilterLink.ts
```diff
@@ -0,0 +1,53 @@
...
+      ],
+      operation.query
+    );
+    if (!clientDirectivesPresent) {
+      return forward(operation);
+    } else {
+
+      checkDocument(operation.query);
+      const newDoc = removeDirectivesFromDocument(
+        [
```
wtrocki: [Optional] From maintenance point of view it will be cool to have array of directives defined somewhere that people can just add when they define directive. It will be easier to maintain as currently new client directive will require changes in couple places in this link.

packages/sync/test/LocalDirectiveFilterLink.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  it("ensures online only directive does not exist after local link", () => {
+    expect(hasDirectives(["onlineOnly"], requestWithOnlineDirective.query));
+    execute(link, requestWithOnlineDirective).subscribe({});
+    expect(testLink.operations.length).equal(1);
+    expect(!(hasDirectives(["onlineOnly"], testLink.operations[0].query)));
+  });
+
+  it("ensures no squash directive does not exist after local link", () => {
```
wtrocki: [Optional] It will be cool to see if it's not crashing when both are applied.

packages/sync/src/links/LocalDirectiveFilterLink.ts
wtrocki: Yeah, but any reason why we do this here? At this point document wasn't touched at all so we kinda duplicating validation that happened before?

packages/sync/src/links/LocalDirectiveFilterLink.ts
```diff
@@ -0,0 +1,47 @@
...
+  Operation
+} from "apollo-link";
+import { hasDirectives, removeDirectivesFromDocument, checkDocument } from "apollo-utilities";
+import { LocalDirectivesArray, MUTATION_QUEUE_LOGGER, LocalDirectives } from "../config/Constants";
+import debug from "debug";
+
+export const logger = debug(MUTATION_QUEUE_LOGGER);
+
+export class LocalDirectiveFilterLink extends ApolloLink {
+  private connectionRemoveOnlineOnly = {
```
wtrocki: If we do not use array to build those then this array makes no sense as there is still manual input needed.

PR #181, opened 2018-12-07
Expose client type

wtrocki:
## Motivation

Add missing return type for client.
Map existing complex type into simple one to simplify import on client side.


darahayes:
@wtrocki code changes look fine to me but I'm just wondering about the sub module?

wtrocki:
Ah.. it's happening all the time! My bad!. Removing it

PR #184, opened 2018-12-11
Fix NetworkStatus.isOffline() on Cordova/iOS

danielpassos:


packages/sync/src/offline/CordovaNetworkStatus.ts
```diff
@@ -19,6 +15,6 @@ export class CordovaNetworkStatus implements NetworkStatus {
   }
 
   public isOffline(): boolean {
-    return (navigator.connection.type === this.NONE);
```
wtrocki: I recognize this change, but I would love to test this carefully and see how this will interact with deviceReady

wtrocki:
@danielpassos I think we need to replicate this with promise.

wtrocki:
need change of usage in queue

packages/sync/src/links/OfflineQueueLink.ts
```diff
@@ -135,11 +135,14 @@ export class OfflineQueueLink extends ApolloLink {
   public openQueueOnNetworkStateUpdates(): void {
     const self = this;
     if (this.networkStatus) {
-      if (this.networkStatus.isOffline()) {
-        this.close();
-      } else {
-        this.open();
-      }
+      this.networkStatus.isOffline().then(offline => {
```
wtrocki: I believe that we need to set initial state before we restore queue.
Happy to do that in other PR.

wtrocki:
@danielpassos What is the story here?

danielpassos:
@wtrocki That is exactly that make https://github.com/aerogear/apollo-voyager-ionic-example/pull/7 works. Not sure if I got you

PR #185, opened 2018-12-11
Add offline queue listener

wtrocki:
## Motivation
 
Allow to detect operations on queue in order to provide feedback for users and their UI. 
This simple listener can be extended over the time. For the moment we have:

- Detection of the element added to queue
- Detection of elements flushed from queue

This code will still be vulnerable to issues with RestoreHandler. 
I'm going to deal with this in separate PR as it's not as trivial as actual listener.
If you guys have idea about other methods that can be exposed let me know.

Verification will be possible on separate PR for ionic sample app- link soon.

## Notes

Since there is no implementation on our side we do not have unit tests for this particular feature.

## Verification

1. Please run example app using
https://github.com/aerogear/apollo-voyager-ionic-example/pull/5

2. Go offline
3. Add new item
4. See queue going to be zero

### DIFF
--- a/packages/sync/src/links/OfflineQueueLink.ts
+++ b/packages/sync/src/links/OfflineQueueLink.ts
@@ -14,6 +14,7 @@ import { NetworkStatus, NetworkInfo } from "../offline";
 import { DataSyncConfig } from "../config";
 import { squashOperations } from "../offline/squashOperations";
 import debug from "debug";
+import { OfflineQueueListener } from "../offline/OfflineQueueListener";
 
 export const logger = debug(MUTATION_QUEUE_LOGGER);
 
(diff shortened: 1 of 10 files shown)


packages/sync/src/links/OfflineQueueLink.ts
```diff
@@ -67,6 +70,9 @@ export class OfflineQueueLink extends ApolloLink {
       forward(operation).subscribe(observer);
     });
     this.opQueue = [];
+    if (this.listener && this.listener.queueCleared) {
```
danielpassos: Sorry for the idiot question but what `this.listener.queueCleared` do in the if/test?

wtrocki: Both listener and individual methods are optional and can be undefined

wtrocki:
Approved again by passos

PR #186, opened 2018-12-12
Optimistic responses

wtrocki:
## Motivation
 
Add helper for building optimistic responses. As we do not override client we can simplify this process by providing this simple helper and document it. This way we can say we support optimistic responses.


packages/sync/README.md
```diff
@@ -1,19 +1,17 @@
-## AeroGear Services Sync SDK
+# AeroGear Apollo GraphQL Client
```
danielpassos: Description in the `package.json` says: _AeroGear Apollo GraphQL Voyager client_

packages/sync/README.md
```diff
@@ -123,12 +130,58 @@ exampleMutation(...) @noSquash {
...
+
+## Implementing Custom Network Status checks
+
+To use your own custom network checks, implement the [NetworkStatus](https://github.com/aerogear/aerogear-js-sdk/blob/master/packages/sync/src/offline/NetworkStatus.ts)
+ interface which provides two functions;
+
+```javascript
+  onStatusChangeListener(callback: NetworkStatusChangeCallback): void;
+
+  isOffline(): boolean;
```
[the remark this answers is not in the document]
wtrocki: To be added in https://github.com/aerogear/aerogear-js-sdk/pull/184

packages/sync/README.md
```diff
@@ -34,64 +32,73 @@ customLinkBuilder | Enables providing custom Apollo Link for processing requests
...
+with the server state thanks to powerful conflict resolution strategies.
+Please follow chapters bellow for more information.
 
-AeroGear Sync SDK provides offline mutations out of the box. By default it uses localStorage to do this and mutations are stored under the cache key of 'offline-mutations-queue'.
+## Querying local cache
+
+By default client will save all performed query results in the cache.
+Data will be available to be used when application goes offline.
+Queries are cached out of the box based on the type and `id` field.
+When performing mutations that affects some queries users can use `refetchQueries` or `update` fields when performing mutations.
```
StephenCoady: should we show a code example here?

packages/sync/README.md
```diff
@@ -123,12 +130,58 @@ exampleMutation(...) @noSquash {
...
 
-## Designing your types
+## Conflicts
 
-When designing your GraphQL schema types `id` field is always required.
-Library will perform business logic assuming that `id` field will be supplied and returned from server. Without this field some offline functionalities will not work properly.
+TODO
 
-# Contribution guide
+### Conflict strategies
```
StephenCoady: I know you have a TODO, but we should probably mention versions somewhere here.

StephenCoady: Yeah you are right, it does make sense to wait. The only reason I pointed it out was the part discussing conflict resolution strategies doesn't have much context without it. Happy to wait until we know exactly what it will look like though.

PR #192, opened 2018-12-18
Release 0.1.0

wtrocki:
## Motivation
 
Just version bump. Still development release, but dropped alpha sticker so our getting started will work without specifying exact version.


wtrocki:
npm publish
+ @aerogear/datasync-js@0.1.0

PR #196, opened 2019-01-04
Feat: Conflict handling feature

wtrocki:
## Motivation

Change involve couple different things:

- Making implementation more object oriented (however there is still work on the actual link itself)
- Adding pluggable next state handler
- Adding listener for conflicts
- Cleaning up conflict resolver strategies - removing deepmerge

Decided to push that early for feedback while I'm continuing doing end to end testing and client improvements.

## Verification

1. Grab https://github.com/aerogear/apollo-voyager-ionic-example/pull/15
2. Npm link this PR (this was not published)
```
npm link . 
```
On Showcase
```
npm link @aerogear/datasync-js
```
3. Start server following instructions
4. Start client
5. Add new item. 
6. Note version is 1
7. Edit item note version changed to 2
8. Open new incognito tab with cloned version
9. Make change in both tab (notice conflict happening)

Note: that integration is currently using some simple dialog. There is
separate ticket to build good UI for this.


darahayes:
👀

StephenCoady:
@wtrocki when you say "notice conflict happening on both" what do you mean? 

Currently I see the GraphQL error on the window which tries to update outdated data but I see nothing on the window with the newer data. Is this correct?

wtrocki:
My bad - obviously other window will not be aware of conflict (and it should not be aware)

StephenCoady:
Thought as much but thought I was missing something major

packages/sync/src/conflicts/strategies.ts
```diff
@@ -1,19 +1,9 @@
-import deepmerge from "deepmerge";
+import { ConflictResolutionData } from "./ConflictResolutionData";
 
-const diffMergeClientWins = (server: ConflictResolutionData, client: ConflictResolutionData) => {
-    return deepmerge(server, client);
+// Used as default strategy for SDK
+export const diffMergeClientWins = (server: ConflictResolutionData, client: ConflictResolutionData) => {
```
darahayes: Looks like this should have the arguments `(operationName: string, server: ConflictResolutionData, client: ConflictResolutionData)`

Because in `conflictLink.ts` you are calling `resolvedConflict = config.conflictStrategy(operation.operationName, data.serverData, data.clientData);`

I'm seeing a bug right now where when you create a conflict in the ionic example app, the `resolvedConflict` data coming back from `conflictLink.ts` is the server data instead of the client data and then it's doing a `nextState` on the server data which bumps the version to a higher version than on the server. This creates another conflict and it never gets resolved in the app.

packages/sync/src/conflicts/strategies.ts
```diff
@@ -1,19 +1,9 @@
...
+  return client;
 };
-const diffMergeServerWins = (server: ConflictResolutionData, client: ConflictResolutionData) => {
-    return deepmerge(client, server);
+export const diffMergeServerWins = (server: ConflictResolutionData, client: ConflictResolutionData) => {
```
darahayes: Same comment as above

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -15,30 +14,28 @@ export const conflictLink = (config: DataSyncConfig): ApolloLink => {
...
   return onError(({ operation, forward, graphQLErrors }) => {
     const data = getConflictData(graphQLErrors);
-    if (data) {
-      const resolvedConflict = getConflictStrategy()(data, operation.variables);
-      // TODO Notify
-      resolvedConflict.version = data.version;
+    if (data && config.conflictStrategy && config.conflictStateProvider) {
+      let resolvedConflict;
+      if (data.resolvedOnServer) {
+        resolvedConflict = data.serverData;
```
darahayes: I'm not too sure about this logic, so if the server resolved the conflict, we are going to get it here, call `nextState` and then call `forward` on the already resolved conflict right? Am I missing something here?

wtrocki:
Working on fixing both issues.

packages/sync/src/conflicts/strategies.ts
wtrocki: WOW. How this is even possible!
TypeScript oh my TypeScript so even you have betrayed me ;)

packages/sync/src/conflicts/conflictLink.ts
wtrocki: Definitely bug. Fixed that.

wtrocki:
Going to add more unit tests to cover serverside/client resolution

StephenCoady:
@wtrocki seeing an issue where I can't seem to make server changes win. i.e. when there is a conflict the server data is always over written with the new client attempt.

StephenCoady:
> @wtrocki seeing an issue where I can't seem to make server changes win. i.e. when there is a conflict the server data is always over written with the new client attempt.

Nevermind, issue was between keyboard and chair 🤦‍♂️

PR #199, opened 2019-01-09
Data sync protected

aliok:
https://issues.jboss.org/browse/AEROGEAR-8294

### DIFF
--- a/packages/sync/src/config/HeaderProvider.ts
+++ b/packages/sync/src/config/HeaderProvider.ts
@@ -2,5 +2,5 @@
  * Interface for HeaderProvider
  */
 export interface HeaderProvider {
-  getHeaders(): {[index:string]:string};
+  getHeaders(): {[index: string]: string};
 }
--- a/packages/sync/src/config/HeaderProvider.ts
+++ b/packages/sync/src/config/HeaderProvider.ts
@@ -1,0 +1,6 @@
+/**
+ * Interface for HeaderProvider
+ */
+export interface HeaderProvider {
+  getHeaders(): {[index:string]:string};
+}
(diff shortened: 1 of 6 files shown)


packages/sync/src/links/HeadersLink.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  private headerProvider: HeaderProvider;
+
+  constructor(headerProvider: HeaderProvider) {
+    super();
+    this.headerProvider = headerProvider;
+  }
+
+  public request(operation: Operation, forward: NextLink) {
+    if (this.headerProvider) {
```
wtrocki: Redundant.

packages/sync/src/config/HeaderProvider.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Interface for HeaderProvider
+ */
+export interface HeaderProvider {
+  getHeaders(): {[index: string]: string};
```
wtrocki: [Optional]`{[index: string]: string};`  can be defined as type

wtrocki:
We are missing the main part of auth SDK logic that has `refreshToken` done and it's called inside users app.

aliok:
> Looks good to me. The only question is should we have another HeaderProvider implementation that will add the authorisation header for users that are logged in via the auth module?

> We are missing the main part of auth SDK logic that has refreshToken logic and it's called inside users app.

Why don't we leave that to the user? Otherwise I don't see why we're generalizing things as a `HeaderProvider`.
In Passos's work, we had a `requiresAuth` flag which was to make everything automatic if set. In this new case though, an implementation has to be provided and it is not targeted to be made automatic.

However, I don't know the background of why we're changing in approach. @wtrocki  can give more info perhaps.

See the sample app modification here: https://github.com/aerogear/apollo-voyager-ionic-example/commit/55a9a248e45d94a2293bac5bd20f02eac9c9239b#diff-92c003e539499632c72be26749e89db4R27
PS: sample app is not working good at the moment and the code I wrote there is far from being ideal.

PR #212, opened 2019-01-16
Update keycloak library in Auth SDK

wtrocki:
## Motivation

This is really important version bump as it will allow us to adapt new type of flow called `cordova-native`:
https://github.com/keycloak/keycloak/pull/5387/files

This flow enables much easier integration using tabbed interface. We will need to test it within our integrations and release new version to users. 

Huge thanks to @slaskawi for ideas.

### DIFF
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@aerogear/auth",
-  "version": "2.0.0",
+  "version": "2.1.0",
   "description": "JavaScript Auth module for AeroGear services",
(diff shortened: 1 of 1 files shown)


wtrocki:
We need to verify that first and release it to the wild. I actually have alternative idea that we should probably explore better on the list. Going to post update on the list and then we can see how we will aproach this.

darahayes:
@wtrocki Are you happy if we merge this and create a prerelease?

wtrocki:
Yes. Merged!

PR #213, opened 2019-01-21
Enable offline mutations when server is unreachable

jhellar:
### Description

- allow updates on items created while offline
  - when back online, operations in mutations queue are send one by one (waiting for previous one to finish). If there was optimistic id for the operation, and server returns real id, next operations with this optimistic id in queue are updated with this new id
- enable offline mutations when server is unreachable (e.g. down)
  - two more links were added to `compositeQueueLink`
    - `apollo-link-error` - so that offlineQueueLink can determine there was a network error
    - `apollo-link-retry` - this was added so that operation in offline queue is retried until there is no network error
  - for cache updates when replaying offline mutations, `proxyUpdate` function must be provided by user

https://issues.jboss.org/browse/AEROGEAR-8333
https://issues.jboss.org/browse/AEROGEAR-8334
https://issues.jboss.org/browse/AEROGEAR-8345

#### Verification steps

1. `npm link` those changes to `apollo-voyager-ionic-example` app
2. perform next steps in `apollo-voyager-ionic-example` app

For `allow updates on items created while offline` verification:

1. go offline
2. create item then make some updates on it
3. go back online
4. when offline updates are performed you should still see the updated item

For `enable offline mutations when server is unreachable` verification:

1. kill server
2. make some mutations
3. start the server again
4. offline mutations should be processed after short time

#### Checklist

- [x] update ionic example app - https://github.com/aerogear/apollo-voyager-ionic-example/pull/27
- [x] update README
- [x] add verification steps


wtrocki:
I need to check this as soon as possible. I have it in my queue

wtrocki:
@jhellar I quickly checked this change. I would love to spend more time on this to see how it will fit to our global strategy but this looks great. I know that there is another implementation of offline solution:
https://github.com/carvajalconsultants/apollo-link-offline

This implementation totally ignores the fact of offline/queue stuff and enqueue requests on failed network connections. This way it includes both offline/online state and also failed server requests. 

We may do quick workshop about this with the team to get the consensus over the approaches that are possible and integrate this PR in the master with the sense that we have clear path for supporting offline cases. Thank you so much for effort put into this! I'm pretty sure that we might slightly change our aproach and go with the way you have proposed.

jhellar:
@wtrocki thanks for looking at this and for the link - if I understand it right the apollo-link-offline retries the mutations on network_errors after some time? If so our current approach (watching for online/offline state) has one advantage in this - it will send the offline mutations immediately once the device is back online. There are more pros/cons for both approaches.

+1 for having quick workshop about this

wtrocki:
I think we need mixture of both approaches and this is where your PR comes in. It's not trivial feature and it will have many corner cases so I believe is best to list them and see how we can meet them. Pretty confident that this PR is the answer but want to make sure that we are ok and I'm personally not missing anything.

packages/sync/src/VoyagerClient.ts
```diff
@@ -0,0 +1,28 @@
...
+export class VoyagerClient extends ApolloClient<NormalizedCacheObject> {
+  private readonly proxyUpdate?: ProxyUpdate;
+
+  constructor(options: ApolloClientOptions<NormalizedCacheObject>, proxyUpdate?: ProxyUpdate) {
+    super(options);
+
+    this.proxyUpdate = proxyUpdate;
+  }
+
+  public mutate<T, TVariables = OperationVariables>(options: MutationOptions<T, TVariables>): Promise<FetchResult<T>> {
```
wtrocki: We could chat about this. Why we need to wrap mutate. I was hopping to go different way and make sync really lightweight thing and have set of separate packages that extends apollo client. Is there any limitation in actual client?

packages/sync/src/ProxyUpdate.ts
```diff
@@ -0,0 +1,5 @@
+import { MutationUpdaterFn } from "apollo-client";
+
+type ProxyUpdate = (mutationName: string) => MutationUpdaterFn;
```
wtrocki: I see aproach from: https://medium.com/twostoryrobot/a-recipe-for-offline-support-in-react-apollo-571ad7e6f7f4

packages/sync/src/VoyagerClient.ts
jhellar: yeah, probably better to do this differently. I went with approach from the article you linked in your other comment. This way it is not necessary to specify update functions again for each mutation. Last commit removes the wrapper.

jhellar:
@StephenCoady should be ready for review now

packages/sync/src/VoyagerClient.ts
wtrocki: Oh.. no.. Let's not made changes until we will have idea how this works.

wtrocki:
Update from 1:1 sync. We decided to extract OptimisticMapping feature to separate PR.

PR #215, opened 2019-01-21
DO NOT MERGE enabled tests

StephenCoady:
pr to re-add tests which check offline functionality


wtrocki:
What happened here ?

StephenCoady:
I was only using this as I naively thought I had fixed the problem. The tests I was trying to fix here are resolved by @jhellar 's PR. I am run locally for now but ignore this branch.

PR #216, opened 2019-01-22
moved directive, added tests

StephenCoady:
fixed directives and added a test case


wtrocki:
What happened here? :>

StephenCoady:
@wtrocki I had forgotten I had this PR, but we found out the bug was a problem with the test suite and not the code. @jhellar is updating the tests separately.

PR #218, opened 2019-01-22
fix: Move core to be in peerDependencies

wtrocki:
## Motivation

Core package is used as singleton and when having more than single instance of core app config will stop working. While this is only temporary fix for the invalid architecture it will provide cleaner way to manage core<> app dependency.

How things will work now:
- App will be required to configure entire platform and setup metrics
- Each individual package will be relying on the core provided in app dependency.

Eventually we can map app package to core and let it be used inside every SDK - however this will affect documentation and end user API so not sure if we are ok with this move.

EDIT:

Actually core.init and app init are the same so we could do that if you guys think this is the way to go.

### DIFF
--- a/packages/security/package.json
+++ b/packages/security/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@aerogear/security",
-  "version": "2.1.0-rc1",
+  "version": "2.1.0-rc2",
   "description": "AeroGear Services JavaScript Security SDK",
   "main": "dist/index.js",
   "types": "types/index.d.ts",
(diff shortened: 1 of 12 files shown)


packages/security/package.json
```diff
@@ -26,12 +26,13 @@
...
+    "@aerogear/core": "2.1.0-rc2",
     "@types/chai": "^4.1.3",
     "chai": "^4.1.2",
     "mocha": "^5.1.1",
     "nyc": "^11.7.1"
   },
-  "dependencies": {
-    "@aerogear/core": "2.1.0-rc1"
+  "peerDependencies": {
+    "@aerogear/core": ">2.0.1"
```
darahayes: ???

packages/sync/package.json
```diff
@@ -44,7 +44,7 @@
     "typescript": "^3.1.6"
   },
   "peerDependencies": {
-    "@aerogear/core": ">2.0.0"
+    "@aerogear/core": ">2.0.1"
```
darahayes: Where is this version number coming from?

wtrocki: Lowest supported version that will match this. Explained that in releng docs.

PR #220, opened 2019-01-23
Allow updates on items created while offline

jhellar:
### Description

https://issues.jboss.org/browse/AEROGEAR-8444

Extracted from https://github.com/aerogear/aerogear-js-sdk/pull/213.

Allow updates on items created while offline. When back online, operations in mutations queue are send one by one (waiting for previous one to finish). If there was optimistic id for the operation, and server returns real id, next operations with this optimistic id in queue are updated with this new id.

This will only work until client is re-initialized - `replayOfflineMutations` does not have this change. https://github.com/jhellar/aerogear-js-sdk/pull/1 enables this also for `replayOfflineMutations`.

#### Verification steps

1. `npm link` those changes to `apollo-voyager-ionic-example` app
2. go offline
3. create item then make some updates on it
4. go back online
5. when offline updates are performed you should still see the updated item

### DIFF
--- a/packages/sync/src/cache/createOptimisticResponse.ts
+++ b/packages/sync/src/cache/createOptimisticResponse.ts
@@ -48,6 +48,7 @@ export const createOptimisticResponse =
     };
     if (addId) {
       optimisticResponse[operation][idField] = generateId();
+      optimisticResponse[operation].__optimisticId = true;
     }
 
     return optimisticResponse;
--- a/packages/sync/src/cache/createOptimisticResponse.ts
+++ b/packages/sync/src/cache/createOptimisticResponse.ts
@@ -1,7 +1,13 @@
 
+const CLIENT_ID_PREFIX = "client:";
+
+export const hasClientGeneratedId = (optimisticResponse: any, operationName: string) => {
+  return optimisticResponse[operationName].id.startsWith(CLIENT_ID_PREFIX);
+};
+
 // Helper method for ID generation ()
 const generateId = (length = 8) => {
-  let result = "client:";
+  let result = CLIENT_ID_PREFIX;
   const chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
 
   for (let i = length; i > 0; i -= 1) {
--- a/packages/sync/src/cache/createOptimisticResponse.ts
+++ b/packages/sync/src/cache/createOptimisticResponse.ts
@@ -1,7 +1,7 @@
 
 // Helper method for ID generation ()
 const generateId = (length = 8) => {
-  let result = "";
+  let result = "client:";
   const chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
 
   for (let i = length; i > 0; i -= 1) {
@@ -48,7 +48,6 @@ export const createOptimisticResponse =
     };
     if (addId) {
       optimisticResponse[operation][idField] = generateId();
-      optimisticResponse[operation].__optimisticId = true;
     }
 
     return optimisticResponse;
--- a/packages/sync/src/links/OfflineQueueLink.ts
+++ b/packages/sync/src/links/OfflineQueueLink.ts
@@ -72,7 +72,6 @@ export class OfflineQueueLink extends ApolloLink {
         forward(operation).subscribe({
           next: result => {
             this.updateIds(opEntry, result);
-            this.storage.setItem(this.key, JSON.stringify(this.opQueue));
             if (observer.next) { observer.next(result); }
           },
           error: error => {
@@ -178,7 +177,7 @@ export class OfflineQueueLink extends ApolloLink {
   ): void {
     const operation = operationEntry.operation;
     const optimisticResponse = operationEntry.optimisticResponse;
-    if (optimisticResponse && optimisticResponse[operation.operationName].__optimisticId) {
+    if (optimisticResponse && optimisticResponse[operation.operationName].id.startsWith('client:')) {
       const optimisticId = optimisticResponse[operation.operationName].id;
       this.opQueue.forEach(({ operation: op }) => {
         if (op.variables.id === optimisticId && result.data) {
--- a/packages/sync/src/links/OfflineQueueLink.ts
+++ b/packages/sync/src/links/OfflineQueueLink.ts
@@ -14,6 +14,7 @@ import { DataSyncConfig } from "../config";
 import { squashOperations } from "../offline/squashOperations";
 import * as debug from "debug";
 import { OfflineQueueListener } from "../offline";
+import { hasClientGeneratedId } from "../cache/createOptimisticResponse";
 
 export const logger = debug.default(MUTATION_QUEUE_LOGGER);
 
@@ -67,22 +68,17 @@ export class OfflineQueueLink extends ApolloLink {
     logger("MutationQueue is open", this.opQueue);
     this.isOpen = true;
     for (const opEntry of this.opQueue) {
-      const { operation, forward, observer } = opEntry;
-      await new Promise(resolve => {
-        forward(operation).subscribe({
-          next: result => {
-            this.updateIds(opEntry, result);
-            if (observer.next) { observer.next(result); }
-          },
-          error: error => {
-            if (observer.error) { observer.error(error); }
-          },
-          complete: () => {
-            if (observer.complete) { observer.complete(); }
-            resolve();
-          }
-        });
-      });
+      let result;
+      try {
+        result = await this.forwardQueuedOperation(opEntry);
+      } catch (error) {
+        // TODO: notify about failed operation via OfflineQueueListener
+        continue;
+      }
+      const { operation: { operationName }, optimisticResponse } = opEntry;
+      if (result && optimisticResponse && hasClientGeneratedId(optimisticResponse, operationName)) {
+        this.updateIds(opEntry, result);
+      }
     }
     this.opQueue = [];
     if (this.listener && this.listener.queueCleared) {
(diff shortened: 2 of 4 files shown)


wtrocki:
Added to my task list. Going to review that today.

danielpassos:
👀

wtrocki:
👀

packages/sync/src/links/OfflineQueueLink.ts
```diff
@@ -62,12 +63,28 @@ export class OfflineQueueLink extends ApolloLink {
     this.operationFilter = filter;
   }
 
-  public open() {
+  public async open() {
```
wtrocki: I believe that with this change and entire thing being async now we can get rid or refactor `OfflineRestoreHandler`.

packages/sync/src/links/OfflineQueueLink.ts
```diff
@@ -62,12 +63,28 @@ export class OfflineQueueLink extends ApolloLink {
...
-    this.opQueue.forEach(({ operation, forward, observer }) => {
-      forward(operation).subscribe(observer);
-    });
+    for (const opEntry of this.opQueue) {
+      const { operation, forward, observer } = opEntry;
+      await new Promise(resolve => {
+        forward(operation).subscribe({
+          next: result => {
+            this.updateIds(opEntry, result);
+            this.storage.setItem(this.key, JSON.stringify(this.opQueue));
```
wtrocki: I believe that this is could be the most controversial change. If you have 50 edits in queue for this change we will Parse and Stringify entire queue so many times significantly affecting performance of the application when starting (this will usually happen on startup). To prevent from that we should:

- refactor OfflineRestoreHandler to be part of this so it will be separated from regular processing.
- Avoid setItem being called here as opQueue as we really keep everything in memory and all requests should be processed within this loop.
- Clear out async story - this method is async but we do not do await bellow when calling open etc.

First one can be done after this is merged.

packages/sync/src/links/OfflineQueueLink.ts
```diff
@@ -153,4 +171,21 @@ export class OfflineQueueLink extends ApolloLink {
...
     }
   }
+
+  private updateIds(
+    operationEntry: OperationQueueEntry,
+    result: FetchResult
+  ): void {
+    const operation = operationEntry.operation;
+    const optimisticResponse = operationEntry.optimisticResponse;
+    if (optimisticResponse && optimisticResponse[operation.operationName].__optimisticId) {
```
wtrocki: [Idea] Instead of flag we can simply have `__optimisticId` or `__clientId` field. This will be more beneficial because at any point we will be able to get both server and client id.

packages/sync/src/links/OfflineQueueLink.ts
```diff
@@ -153,4 +171,21 @@ export class OfflineQueueLink extends ApolloLink {
...
+      const optimisticId = optimisticResponse[operation.operationName].id;
+      this.opQueue.forEach(({ operation: op }) => {
+        if (op.variables.id === optimisticId && result.data) {
```
wtrocki: I think that if we go with this to avoid id clashes between server and client we need to generate ID's using `-` or with some prefix like `client:3gj4hgj34hbgdfgf`

packages/sync/src/links/OfflineQueueLink.ts
```diff
@@ -153,4 +171,21 @@ export class OfflineQueueLink extends ApolloLink {
...
+          op.variables.id = result.data[operation.operationName].id;
+        }
+      });
+      this.storage.setItem(this.key, JSON.stringify(this.opQueue));
```
wtrocki: This is duplicate (we do that already outside this method)

packages/sync/src/links/OfflineQueueLink.ts
danielpassos: @wtrocki That is a very good catch. I think a prefix is a good solution

packages/sync/src/links/OfflineQueueLink.ts
jhellar: The reason for saving the queue back to store is that if the app is closed (or page refreshed) during processing, the changes would not be persisted. The worst thing that could happen is that some updates in queue could later fail because of ids that have not been updated. But I think that if we can easily prevent it we probably should even though the performance would be worse.

Also we do the same currently on master - once there is a response from server, operation is removed from queue and queue is saved to store (`cancelOperation`). Now we do the save twice - when there is a response from server, then when ids are updated. Maybe this could be somehow improved?

Is stringifying that slow? There is still communication with server for every operation, which IMHO will be slower compared to saving queue to store.

packages/sync/src/links/OfflineQueueLink.ts
jhellar: Yes, I think that if we do this, we can also remove the `__optimisticId` flag.

packages/sync/src/links/OfflineQueueLink.ts
jhellar: ok, I overlooked this, will fix it

jhellar:
@wtrocki thanks for review. It was easier to separate the rest of the original PR in way that it depends on this PR, so it is here https://github.com/jhellar/aerogear-js-sdk/pulls. Some of the things you mentioned in your comments are in some way done there - e.g. `OfflineRestoreHandler` just pushes mutations back to queue.

packages/sync/src/links/OfflineQueueLink.ts
wtrocki: > The reason for saving the queue back to store is that if the app is closed (or page refreshed) during processing, the changes would not be persisted. 

Hmm... so maybe OfflineRestoreHandler should be triggered by user then at the moment they want to push the changes. I feel like this is really an corner case - if someone refresh page during the sending request to server it will have the same effect.

> The worst thing that could happen is that some updates in queue could later fail because of ids that have not been updated.

After getting thru the changes I think in this case queue will be empty already.

jhellar:
@wtrocki please create jiras for those bugs:

1. fix timing issues around async/await
2. reduce number of writes to store for offline mutations
3. add ID replacement to offlineRestoreHandler

I will simplify optimisticID to differentiate between client/server side IDs.

wtrocki:
Working on it!

wtrocki:
IDEA: https://github.com/aerogear/aerogear-js-sdk/commit/224504997e946232e5cf8660b9cf617e43f03a14

jhellar:
@wtrocki I've removed the `__optimisticId` flag. Now there is `client:` prefix for every optimistic id.

wtrocki:
Great! I have it on my radar to verify it and merge it today.
Stay tunned.

wtrocki:
@jhellar "add ID replacement to offlineRestoreHandler" will not be valid If we follow aproach of having closed queue. Then id replacement will happen only in link.

Created epic: https://issues.jboss.org/browse/AEROGEAR-8439
Feel free to throw more stuff into this.

packages/sync/src/cache/createOptimisticResponse.ts
```diff
@@ -1,7 +1,7 @@
 
 // Helper method for ID generation ()
 const generateId = (length = 8) => {
-  let result = "";
+  let result = "client:";
```
wtrocki: This will look better as  constant somewhere above.

packages/sync/src/links/OfflineQueueLink.ts
```diff
@@ -62,12 +63,27 @@ export class OfflineQueueLink extends ApolloLink {
...
-  public open() {
+  public async open() {
     logger("MutationQueue is open", this.opQueue);
     this.isOpen = true;
-    this.opQueue.forEach(({ operation, forward, observer }) => {
-      forward(operation).subscribe(observer);
-    });
+    for (const opEntry of this.opQueue) {
+      const { operation, forward, observer } = opEntry;
+      await new Promise(resolve => {
```
wtrocki: Better to extract this to separate method: `updateId's` to make this more maintainable.

packages/sync/src/links/OfflineQueueLink.ts
```diff
@@ -153,4 +170,21 @@ export class OfflineQueueLink extends ApolloLink {
...
     }
   }
+
+  private updateIds(
+    operationEntry: OperationQueueEntry,
+    result: FetchResult
+  ): void {
+    const operation = operationEntry.operation;
+    const optimisticResponse = operationEntry.optimisticResponse;
+    if (optimisticResponse && optimisticResponse[operation.operationName].id.startsWith("client:")) {
```
wtrocki: Bad habbit of leaking implementations. We have two dangling flags in the code: 
`client:` is in 2 classes. Let's make separate method in optimisticResponse file and use it here.

wtrocki:
@jhellar Played with this stuff and absolutely love this. Added some minor comments after that I think we are ready to:

- bump package and release dev version.
- Integrate that with sample app
- Do more more testing
- Merge all :)

packages/sync/src/cache/createOptimisticResponse.ts
jhellar: should be fixed now

packages/sync/src/links/OfflineQueueLink.ts
jhellar: I've created `forwardQueuedOperation` method [1] which basically makes possible to await `forward`.

[1] https://github.com/aerogear/aerogear-js-sdk/pull/220/files#diff-334f3fa2eb415fbeb3824ef3b38b5580R115

packages/sync/src/links/OfflineQueueLink.ts
jhellar: should be fixed now

packages/sync/src/links/OfflineQueueLink.ts
```diff
@@ -62,12 +63,27 @@ export class OfflineQueueLink extends ApolloLink {
...
     this.isOpen = true;
-    this.opQueue.forEach(({ operation, forward, observer }) => {
-      forward(operation).subscribe(observer);
-    });
+    for (const opEntry of this.opQueue) {
+      const { operation, forward, observer } = opEntry;
+      await new Promise(resolve => {
+        forward(operation).subscribe({
+          next: result => {
+            this.updateIds(opEntry, result);
```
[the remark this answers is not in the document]
jhellar: should be fixed now

wtrocki:
@jhellar Awesome job! Let's bump version to x.x.x-dev.1.
I will publish that and update example app in order to test this.

PR #223, opened 2019-01-25
 Add Auth header provider

danielpassos:


packages/auth/package.json
```diff
@@ -49,7 +49,6 @@
     "branches": 80
   },
   "devDependencies": {
-    "@aerogear/core": "2.1.0-rc2",
```
wtrocki: Core needs to stay as we cannot build without it

danielpassos: I have push force removing the commit who remove it

PR #225, opened 2019-01-28
Refactored integration tests for sync

jhellar:
### Description

https://issues.jboss.org/browse/AEROGEAR-8111

Refactored integration tests so that test cases don't depend on each other.

### DIFF
--- a/packages/sync/integration_test/utils/graphql.queries.js
+++ b/packages/sync/integration_test/utils/graphql.queries.js
@@ -40,16 +40,16 @@ export const UPDATE_TASK = gql`
 `;
 
 export const ONLINE_ONLY = gql`
-  mutation onlineOnly($id: ID!) @onlineOnly {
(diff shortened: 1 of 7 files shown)


packages/sync/integration_test/utils/graphql.queries.js
```diff
@@ -40,16 +40,16 @@ export const UPDATE_TASK = gql`
 `;
 
 export const ONLINE_ONLY = gql`
-  mutation onlineOnly($id: ID!) @onlineOnly {
-    onlineOnly(id: $id) {
+  mutation onlineOnly($id: ID!) {
+    onlineOnly(id: $id) @onlineOnly {
```
wtrocki: Is this test enabled now?

jhellar: No, for some reason it still fails. I've created item for it in test plan, so that this is tested manually.

PR #226, opened 2019-01-30
Refactor offlineRestoreHandler

jhellar:
### Description

https://issues.jboss.org/browse/AEROGEAR-8445

When replaying offline mutations in OfflineRestoreHandler, offline queue should be closed, so that all mutations are back in queue. Then queue can be opened depending on network state. This way IDs will be updated also when replaying mutations.


packages/sync/src/createClient.ts
```diff
@@ -30,6 +30,7 @@ export const createClient = async (userConfig?: DataSyncConfig): Promise<Voyager
     storage,
     clientConfig.mutationsQueueName);
   await offlineMutationHandler.replayOfflineMutations();
+  offlineQueueLink.openQueueOnNetworkStateUpdates();
```
wtrocki: [Idea] Add setTimeout with 3/5 seconds delay?
If not maybe we can await?

packages/sync/src/offline/OfflineRestoreHandler.ts
```diff
@@ -41,28 +42,28 @@ export class OfflineRestoreHandler {
...
+      this.apolloClient.mutate({
+        variables: item.operation.variables,
+        mutation: item.operation.query,
+        optimisticResponse: item.optimisticResponse
+      });
+    });
 
-    // wait before it was cleared
-    await this.clearOfflineData();
+    await this.waitForOperationsEnqueued();
```
wtrocki: this works but I think we can simply use listener for enqueued operations and await for this.

wtrocki:
Verification:
- enqueue large amount of offline data
go offline
- restart app
- note that offline queue restore handler processed all requests
- go online 
- see queue cleared

packages/sync/src/createClient.ts
jhellar: Is it to wait for queue being opened and all operations sent, so that user won't send operation while queue is being processed? This could happen also later whenever is queue opened.  To prevent this, can we enqueue all incoming operations while queue is being processed?

packages/sync/src/offline/OfflineRestoreHandler.ts
jhellar: Was also thinking about that, but not really sure how to do that - user provides his own listener via `clientConfig`. So should we wrap it, so that we can listen for events as well?

jhellar:
@wtrocki thanks for review, will add test for the verification steps.

packages/sync/src/createClient.ts
wtrocki: https://github.com/aerogear/aerogear-js-sdk/blob/master/packages/sync/src/links/OfflineQueueLink.ts#L110

Tells me that you getting notification when things are enqueued.

wtrocki: so it will be something like:
```javascript
new Promise(res,rej){
  mutate(..)
  listener(()=>{res()})
```
}

wtrocki:
@jhellar Awesome! I will also play with this manually but this change makes things much more cleaner and simple.

wtrocki:
@jhellar It works like dream!

StephenCoady:
👀

PR #227, opened 2019-01-30
Remove customLinkBuilder

jhellar:
### Description

https://issues.jboss.org/browse/AEROGEAR-8452

Links like offlineQueueLink or conflictLink should be exposed, so that users can build their own link chains. For example user may want to add retry link before http link. It should be documented how to build custom chains.

### DIFF
--- a/packages/sync/src/index.ts
+++ b/packages/sync/src/index.ts
@@ -4,3 +4,4 @@ export * from "./conflicts";
 export * from "./links";
 export * from "./offline";
 export * from "./cache/createOptimisticResponse";
+export * from "./utils/helpers";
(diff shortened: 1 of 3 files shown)


wtrocki:
@jhellar They will be eventually in separate packages. THen entire client will be really thin.
I was proponent to remove fact that people can supply their custom link as this may not be as usable

jhellar:
@wtrocki ok. Other reason I created this PR is this https://github.com/aerogear/aerogear-js-sdk/blob/master/packages/sync/src/links/LinksBuilder.ts#L13

I think it would be better not to pass the `offlineQueueLink` like this. Would it make sense to remove the `readme` and `expose` parts from this PR and leave just the restructuring, so that `offlineQueueLink` is accessible in `createClient` where it is then used?

wtrocki:
I would say entire thing is not needed. We should remove https://github.com/aerogear/aerogear-js-sdk/blob/master/packages/sync/src/links/LinksBuilder.ts#L18 
Remove this from config and docs and then do everything including restore handler together. 
This will give better flexibility and remove amount of hacks. 

If someone wants to go custom they will just copy or extend this as there is no other implementation.

jhellar:
@wtrocki ok, should be fixed now

jhellar:
@wtrocki the definitions should be exposed now

PR #230, opened 2019-01-31
BREAKING: Remove instance from core

wtrocki:
## Motivation

Fix api on js-sdk.

### Migration:
Users need to pass created app to each service instance

```
const app = init(config)
const auth = new AuthService(app)
```

## Progress

- [x] Migrate new api
- [ ] Fix unit tests
- [ ] Update documentation


packages/app/src/index.ts
```diff
@@ -1,11 +1,56 @@
...
+   * @param id - unique id of the service
+   */
+  public getConfigById(id: string): ServiceConfiguration | undefined {
+    if (this.configurations) {
+      return find(this.configurations, service => {
+        return !!service.id && service.id.toLowerCase() === id.toLowerCase();
+      });
+    }
+    console.error("Configuration not initialized.");
+  }
```
wei-lee: if we let each service to use `ServiceConfiguration` in the constructor, then we don't need to duplicate those methods in the `app` module here.

packages/auth/src/Auth.ts
```diff
@@ -14,8 +15,8 @@ export class Auth {
   private auth: KeycloakInstance;
   private internalConfig: any;
 
-  constructor() {
-    const configuration = INSTANCE.getConfigByType(Auth.TYPE);
+  constructor(app: AeroGearApp) {
```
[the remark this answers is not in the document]
wtrocki: Yes. I had that done in the first iteration but `SecurityService` need to get access to metrics.
Initially I tried to workaround this, but it may be actually valuable to have app as can give us flexibility.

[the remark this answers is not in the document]
wtrocki: WDYT?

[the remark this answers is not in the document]
wei-lee: right. In this case, I say let's just pass in the Metrics instance to the constructor of Security service. This way it will make it very obvious that Security service is depending on the metrics service, but not the `app`.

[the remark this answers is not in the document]
wtrocki: Basing on my understanding it may look like this:
`
const app = new App()
const authConfig = app.getConfigByType(Auth.Type)
const auth = new Auth(authConfig)
`
I can do that but it seems too verbose to me. Considering that config is already known by user

packages/app/src/index.ts
```diff
@@ -1,11 +1,67 @@
 
-import { AeroGearConfiguration, INSTANCE } from "@aerogear/core";
+import find from "lodash.find";
+import console from "loglevel";
+import { AeroGearConfiguration, ServiceConfiguration, MetricsService } from "@aerogear/core";
+
+/**
+ * Abstraction for AeroGear configuration
+ */
+export class ConfigurationService {
```
wei-lee: can we move this to be part of the core module? This way this service module will not need to depend on the `app` module anymore, but just the `core` module. This way the `core` module is the only command module that is shared across all the service modules. All the services module will then be decoupled from the `app` module completely.

wtrocki:
Ok.. So initial idea was that app will control everything, but after recommendation I think we got really clear architecture:

- Core is stateless and just holds implementation for config and 
- App is stateful and used to hold state of the metrics and core. 
- Individual packages depend on core only and get injected all elements

I have tried that with Ionic sample app and it will work like dream. 
So technically architecture will look like this:
screen shot 2019-02-01 at 11 50 51 am

packages/auth/package.json
```diff
@@ -49,7 +49,7 @@
     "branches": 80
   },
   "devDependencies": {
-    "@aerogear/core": "2.1.0-rc2",
+    "@aerogear/core": "2.1.0",
```
wei-lee: should this be in `dependencies` rather than `devDependencies`?

packages/push/package.json
```diff
@@ -28,13 +28,14 @@
   },
   "homepage": "https://github.com/aerogear/aerogear-js-sdk#readme",
   "devDependencies": {
-    "@aerogear/core": "2.1.0-rc2",
+    "@aerogear/core": "2.1.0",
```
wei-lee: this is already declared in `dependencies`

packages/auth/package.json
wtrocki: I put core into peerDependencies for runtime and devDependencies for typescript builds.
We can put core to be in dependencies for each package as form of simplification. 
WDYT?

packages/security/package.json
```diff
@@ -26,12 +26,15 @@
...
     "@types/chai": "^4.1.3",
     "chai": "^4.1.2",
     "mocha": "^5.1.1",
     "nyc": "^11.7.1"
   },
+  "dependencies": {
+    "@aerogear/core": "2.1.0"
+  },
   "peerDependencies": {
     "@aerogear/core": ">2.0.1"
```
wei-lee: we probably can remove this one now, right?

packages/security/package.json
```diff
@@ -26,12 +26,15 @@
   },
   "homepage": "https://github.com/aerogear/aerogear-js-sdk#readme",
   "devDependencies": {
-    "@aerogear/core": "2.1.0-rc2",
+    "@aerogear/core": "2.1.0",
```
wei-lee: this is already declared in `dependencies`

packages/sync/package.json
```diff
@@ -31,7 +31,7 @@
     "access": "public"
   },
   "devDependencies": {
-    "@aerogear/core": "2.1.0-rc2",
+    "@aerogear/core": "2.1.0",
```
wei-lee: same here

packages/auth/package.json
wtrocki: It actually made no difference now but having peerDependencies on those will always make sure that there is this single core there. Core then will come from app module that requires that. However this is more complex idea so maybe let's just put core into dependencies and call it a day. 
Doing this now.

docs/templates/package.json
```diff
@@ -54,7 +55,7 @@
     "ts-node": "5.0.1",
     "typescript": "^2.8.1"
   },
-  "dependencies": {
+  "peerDependencies": {
     "@aerogear/core": "0.1.0"
```
wei-lee: I think `dependencies` is probably right? Do we need it in `devDependencies` as well?

packages/auth/package.json
wei-lee: +1

wtrocki: Ah.. our scripts for version bump putting core into dependencies that is why it landed there. Yes another good reason to not change this.

wtrocki:
Added fixes. @StephenCoady could you take look if you do not have any concerns over the decisions we have made here.

wtrocki:
Current architecture: 
js-sdk architecture

wtrocki:
Going to do another prerelease now.

Issue #198, opened 2019-01-08
Allow to build conflict resolvers objects rather than passing current config.


wtrocki:
## Description
 
Currently conflict resolution is implemented as global function that can be customized by checking different mutation names. We could change that so users will provide object with names of mutations and specific conflict resolutions that needs to happen.

PR #232, opened 2019-02-05
Added ability to pass a resolver object

ziccardi:
# Motivation
Issue: [AEROGEAR-8382](https://issues.jboss.org/browse/AEROGEAR-8382) - #198 

# Description

- Added a new `ConflictResolutionStrategies` interface that allows passing pairs of `operationName` and `resolver` function.
- Added ability to work with both the old way (one function for all the operations) and the new way (one function for each operation)

# Activity

- [x] Add the new function
- [x] Update the readme
- [x] Update doc in comments

### DIFF
--- a/packages/sync/src/config/DataSyncConfig.ts
+++ b/packages/sync/src/config/DataSyncConfig.ts
@@ -1,4 +1,4 @@
-import { ConflictResolutionStrategy, IResolver } from "../conflicts/ConflictResolutionStrategy";
+import { ConflictResolutionStrategy, ConflictResolutionStrategies } from "../conflicts/ConflictResolutionStrategy";
 import { PersistedData, PersistentStore } from "../PersistentStore";
 import { NetworkStatus } from "../offline";
 import { OfflineQueueListener } from "../offline";
@@ -62,11 +62,6 @@ export interface DataSyncConfig {
   auditLogging?: boolean;
 
   /**
-   * The conflict resolution strategy your client should use. By default it takes client version
-   */
-  conflictStrategy?: ConflictResolutionStrategy;
-
-  /**
    * Interface that defines how object state is progressed
    */
   conflictStateProvider?: NextState;
@@ -81,5 +76,8 @@ export interface DataSyncConfig {
    */
   openShiftConfig?: ConfigurationService;
 
-  resolvers?: IResolver;
+  /**
+   * The conflict resolution strategy your client should use. By default it takes client version.
+   */
+  conflictStrategy?: ConflictResolutionStrategy | ConflictResolutionStrategies;
 }
--- a/packages/sync/src/config/DataSyncConfig.ts
+++ b/packages/sync/src/config/DataSyncConfig.ts
@@ -1,4 +1,4 @@
-import { ConflictResolutionStrategy } from "../conflicts/ConflictResolutionStrategy";
+import { ConflictResolutionStrategy, IResolver } from "../conflicts/ConflictResolutionStrategy";
 import { PersistedData, PersistentStore } from "../PersistentStore";
 import { NetworkStatus } from "../offline";
 import { OfflineQueueListener } from "../offline";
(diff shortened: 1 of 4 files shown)


packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -40,7 +40,15 @@ export const conflictLink = (config: DataSyncConfig): ApolloLink => {
         }
       } else {
         // resolve on client
-        resolvedConflict = config.conflictStrategy(operation.operationName, data.serverState, data.clientState);
+        if (config.resolvers && config.resolvers[operation.operationName]) {
+          config.resolvers[operation.operationName](data.serverState, data.clientState)
+        }
+
+        // use global function, if present
+        if (config.conflictStrategy) {
```
ziccardi: @wtrocki Do we have to keep this for compatibility with the previous code?

wtrocki: No.. This is relatively low priority change but we need  to change API before it hits users.

packages/sync/src/config/DataSyncConfig.ts
```diff
@@ -80,4 +80,6 @@ export interface DataSyncConfig {
    * OpenShift specific configuration
    */
   openShiftConfig?: ConfigurationService;
+
+  resolvers?: IResolver;
```
wtrocki: What is that? Missing docs.

wtrocki: Ah.. resolvers are actual name for graphql resolvers on server.
Howe about rename this to conflictHandlers

ziccardi: 👍

packages/sync/src/conflicts/ConflictResolutionStrategy.ts
```diff
@@ -10,3 +10,8 @@ import { ConflictResolutionData } from "./ConflictResolutionData";
  */
 export type ConflictResolutionStrategy =
   (operationName: string, server: ConflictResolutionData, client: ConflictResolutionData) => ConflictResolutionData;
+
+export interface IResolver {
```
wtrocki: Missing docs. 
All end user objects needs to be documented.
Wrong folder (it should be in api)

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -24,7 +24,7 @@ export const conflictLink = (config: DataSyncConfig): ApolloLink => {
 
   return onError(({ response, operation, forward, graphQLErrors }) => {
     const data = getConflictData(graphQLErrors);
-    if (data && config.conflictStrategy && config.conflictStateProvider) {
+    if (data && (config.conflictStrategy || config.resolvers) && config.conflictStateProvider) {
```
wtrocki: IMHO having two ways to handle conflicts is really confusing and also introduces some complexity. We do not need to introduce another config - just support alternative way of configuration on the same config key.

packages/sync/src/conflicts/conflictLink.ts
ziccardi: 👍 ok, will remove that then

wtrocki:
Looks really really nice. I will add more docs and try to simplify it. 
Great job @ziccardi 

In terms of verification I believe that we can add automation tests to existing karma tests in this repository

packages/sync/src/config/DataSyncConfig.ts
wtrocki: How about we still alow people to pass global and IResolver under the same name.
We can use pipe 

`ConflictResolutionStrategy | ConflictResolutionStrategies`

packages/sync/src/conflicts/ConflictResolutionStrategy.ts
wtrocki: Rename to `ConflictResolutionStrategies` or something that will clearly state what is about

wtrocki:
@ziccardi Great work! Code looks amaizing. Let's update readme.md file that contains reference to this part and we are done. Leaving karma testing to your consideration.

LGTM - but going to approve once we bump version.

wtrocki:
@ziccardi Unblocked. Rebase needed.

PR #233, opened 2019-02-06
[WIP] Treat server being down as client being offline

jhellar:
### Description

https://issues.jboss.org/browse/AEROGEAR-8447

##### Refactored offline mutations

`offlineQueueLink` was split into `serialLink` and `networkStateLink`. SerialLink makes sure there is one operation in flight at any time except from operations marked with `parallel` directive. It contains queue where operations wait. This queue is being saved, similarly to offlineQueueLink. networkStateLink blocks operations if offline.

##### Checklist

- [x] split `offlineQueueLink` into `serialLink` and `networkStateLink`
- [ ] add `networkErrorLink`
- [ ] unit tests
- [ ] integration tests


wtrocki:
@jhellar It will be cool to get motivation. While changes are rock solid we need to have some serious concerns to actually serialize graphql queries as this could have huge impact on UI rendering

jhellar:
Main motivation for serializing mutations was that when going back online, it won’t be possible to perform mutations operating on client generated IDs. I can see now that it might not be a good approach.

Possible improvement 1: serialize mutations only when offline

Possible improvement 2: serialize only mutations operating on same client generated IDs - having something like parallel queues where for mutations operating on same client generated ID, separate queue would be created

@wtrocki wdyt?

wtrocki:
> Possible improvement 1: serialize mutations only when offline

Yes. We should not serialized anything non offline. 
If we go back to original Offline implementation - intention was to just simply persist the queries that happened when offline

jhellar:
@wtrocki I've created separate PR for this https://github.com/aerogear/aerogear-js-sdk/pull/237 - removed serialLink and added retryLink. Closing this.

PR #235, opened 2019-02-06
fix: Update depdencies in the sync

wtrocki:
This is going to fix build and allow to merge @ziccardi changes.


wtrocki:
@StephenCoady Still waiting for server release. There will be some changes from @darahayes that will revert breaking changes on conflict and then we can release stuff.

wtrocki:
Outdated

PR #240, opened 2019-02-08
e2e subscriptions

StephenCoady:
## Tasks

Related to: 
https://github.com/aerogear/ionic-showcase/pull/50 <--- verification steps
https://github.com/aerogear/voyager-server/pull/105

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -118,16 +118,21 @@ export class Auth {
   }
 
   /**
-   * Provides an Authorization Bearer header token
+   * Provides an Authorization Context which contains a token and a prebuilt header
    * @param tokenUpdateTime time (in seconds) to refresh the token
    */
-  public getHeaderProvider(tokenUpdateTime: number = 10) {
+  public getAuthContextProvider(tokenUpdateTime: number = 10) {
     return () => {
       const tokenUpdate = this.extract().updateToken(tokenUpdateTime) as any;
       // Keycloak doesn't use a proper promise. Instead it uses success/error.
       return new Promise((resolve, reject) => {
         tokenUpdate.success(() => {
-          resolve({ "Authorization": "Bearer " + this.auth.token });
+          resolve({
+            header: {
+              "Authorization": "Bearer " + this.auth.token
+            },
+            token: this.auth.token
+          });
         }).error((error: any) => {
           console.info("Cannot update keycloak token", error);
           reject(error);
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -125,7 +125,7 @@ export class Auth {
     return () => {
       const tokenUpdate = this.extract().updateToken(tokenUpdateTime) as any;
       // Keycloak doesn't use a proper promise. Instead it uses success/error.
-      return new Promise((resolve, reject) => {
+      return new Promise<AuthContext>((resolve, reject) => {
         tokenUpdate.success(() => {
           resolve({
             header: {
@@ -140,5 +140,10 @@ export class Auth {
       });
     };
   }
+
+}
+export interface AuthContext {
+  header: any;
+  token?: string;
 
 }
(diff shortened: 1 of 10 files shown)


packages/sync/src/createClient.ts
```diff
@@ -28,7 +28,7 @@ export const createClient = async (userConfig?: DataSyncConfig): Promise<Voyager
   const httpLinks = await defaultHttpLinks(clientConfig);
   let link = ApolloLink.from(httpLinks);
   if (clientConfig.wsUrl) {
-    const wsLink = defaultWebSocketLink({ uri: clientConfig.wsUrl });
+    const wsLink = await defaultWebSocketLink(clientConfig, { uri: clientConfig.wsUrl });
```
wtrocki: I think this await is not needed anymore.

packages/auth/src/Auth.ts
```diff
@@ -118,16 +118,21 @@ export class Auth {
...
    */
-  public getHeaderProvider(tokenUpdateTime: number = 10) {
+  public getAuthContextProvider(tokenUpdateTime: number = 10) {
     return () => {
       const tokenUpdate = this.extract().updateToken(tokenUpdateTime) as any;
       // Keycloak doesn't use a proper promise. Instead it uses success/error.
       return new Promise((resolve, reject) => {
         tokenUpdate.success(() => {
-          resolve({ "Authorization": "Bearer " + this.auth.token });
+          resolve({
```
wtrocki: World class job! Like that we DRY. Could we made this method to return explicit interface. 
This will be cool for api docs :)

packages/sync/src/config/AuthContextProvider.ts
```diff
@@ -0,0 +1,10 @@
+
+export interface AuthContext {
+  header: any;
+  token: string;
+}
+
+/**
+ * Interface for AuthContextProvider
+ */
+export type AuthContextProvider = () => Promise<AuthContext>;
```
wtrocki: Not 100% sure but I think we can avoid having this type and return `Promise<AuthContext>` explicitly above.

packages/sync/src/links/HeadersLink.ts
```diff
@@ -4,10 +4,11 @@ import { DataSyncConfig } from "../config/DataSyncConfig";
 
 export const createHeadersLink = (config: DataSyncConfig): ApolloLink => {
   const asyncHeadersLink = setContext(async (operation, previousContext) => {
-    if (config.headerProvider) {
-      const headers = await config.headerProvider();
+    if (config.authContextProvider) {
+      const { header } = await config.authContextProvider();
+      console.log("HEADER: ", header);
```
wtrocki: Remove this part.

packages/sync/src/config/AuthContextProvider.ts
wtrocki: Ah.. checked code locally. I was wrong

packages/sync/src/config/AuthContextProvider.ts
```diff
@@ -1,7 +1,7 @@
 
 export interface AuthContext {
   header: any;
-  token: string;
+  token?: string;
```
wtrocki: Why token is optional? :>

StephenCoady: No that was a mistake, pushing changes to check token before resolving

packages/sync/src/links/LinksBuilder.ts
```diff
@@ -20,12 +20,8 @@ import { LocalDirectiveFilterLink } from "./LocalDirectiveFilterLink";
 export const defaultHttpLinks = async (config: DataSyncConfig): Promise<ApolloLink[]> => {
   const offlineQueueLink = new OfflineQueueLink(config, "mutation");
   const localDirectiveFilterLink = new LocalDirectiveFilterLink();
-
   let httpLink = new HttpLink({ uri: config.httpUrl, includeExtensions: config.auditLogging }) as ApolloLink;
-  if (config.headerProvider) {
-    httpLink = concat(createHeadersLink(config), httpLink);
-  }
-
+  httpLink = concat(createHeadersLink(config), httpLink);
```
wtrocki: This affects integration tests. Like previous version more ;)
"There is no link if there is no headers you know"

PR #237, opened 2019-02-07
Treat server being down as client being offline

jhellar:
### Description

This is implemented with custom `RetryLink` [1]. It has a queue for all incoming operations. When there is a new operation, it will try to forward it. On network error it will schedule next try. If there is operation with response from server, all operations in queue are force retried.

When creating `RetryLink` user can provide `shouldRetry` function to cancel retry if needed.

[1] https://github.com/aerogear/aerogear-js-sdk/pull/237/files#diff-46acc41354166331ff353344d6065e85

##### Checklist

- [x] added retryLink
- [x] unit tests
- [ ] ~~integration tests~~
- [ ] documentation

Integration tests moved to separate jira - https://issues.jboss.org/browse/AEROGEAR-8521

PR #241, opened 2019-02-12
Aerogear 8447 2

StephenCoady:
replica of [this PR](https://github.com/aerogear/aerogear-js-sdk/pull/237) by @jhellar 

pushing here to preserve his fork.

@wtrocki can you take a look and tell me if the link flow still looks correct, had to do a small bit of refactoring

### Description

This is implemented with custom `RetryLink` [1]. It has a queue for all incoming operations. When there is a new operation, it will try to forward it. On network error it will schedule next try. If there is operation with response from server, all operations in queue are force retried.

When creating `RetryLink` user can provide `shouldRetry` function to cancel retry if needed.

[1] https://github.com/aerogear/aerogear-js-sdk/pull/237/files#diff-46acc41354166331ff353344d6065e85

##### Checklist

- [x] added retryLink
- [x] unit tests
- [ ] ~~integration tests~~
- [ ] documentation

Integration tests moved to separate jira - https://issues.jboss.org/browse/AEROGEAR-8521


wtrocki:
Going to verify this. Check integration tests etc. still some work needed.

wtrocki:
Pushed final version. Released all under: @aerogear/voyager-client@2.1.0-dev.8

wtrocki:
Ok. Maybe to reduce impact and force users to go with retry handler we could just allow users to pass retryFunction. If no retry function is passed then retry link will be disabled. 
This will be much cleaner aproach. 

@psturc @StephenCoady @danielpassos

StephenCoady:
@wtrocki +1 on that approach. covers us from worrying about things like infinite loops too

wtrocki:
😆

wtrocki:
Great work @jhellar! Really good to have this in the project and available to community!

PR #244, opened 2019-02-18
Circleci fixes

wtrocki:
Update some scripts in order to see if build still works


wtrocki:
This needs more work. Closing as there couple ideas that will not work here.

PR #247, opened 2019-02-20
fix: changed link flow to fix local directives

StephenCoady:
### Description
This small change fixes a bug whereby the local only directives were not being removed from the document before being sent to the server.

### Verification
Using this pr as the voyager-client in the showcase app, add @onlineOnly directive or @noSquash directive to the mutations and verify that everything works as expected.

@wtrocki we should probably add some feedback to the app as part of onlineOnly throwing an error in the client (at the moment its just in the console). however it will be hard to differentiate from a normal network error and an intentional one from onlineOnly. any thoughts?

### DIFF
--- a/packages/sync/integration_test/test/offline.test.js
+++ b/packages/sync/integration_test/test/offline.test.js
@@ -24,7 +24,7 @@ const newClient = async (clientOptions = {}) => {
   return await createClient(config);
 };
 
-describe('AeroGear Apollo GraphQL Voyager client', function() {
+describe('AeroGear Apollo GraphQL Voyager client', function () {
 
   this.timeout(1000);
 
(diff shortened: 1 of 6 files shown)


wtrocki:
Pst.. Can be also enable suspended tests and check if they pass now. I do not see reason why they will not pass a the moment.

packages/sync/src/offline/OfflineQueue.ts
```diff
@@ -50,7 +51,7 @@ export class OfflineQueue extends OperationQueue {
   }
 
   protected enqueueEntry(entry: OperationQueueEntry) {
-    if (this.squashOperations) {
+    if (this.squashOperations && !isNotSquashable(entry.operation)) {
```
wtrocki: We have `NotIsNotSquashable` to decode that to real life:
- Squashable (hey we squash things)
- Not (We do not)
- Not Again (So apparently we do)

This is the only usage of this method so maybe we could avoid double negation?
How about having `isSquashable` :)

wtrocki:
After testing I see that we enqueue `Queries` now in local storage.

packages/sync/src/links/LinksBuilder.ts
```diff
@@ -90,8 +90,6 @@ function createOfflineLink(config: DataSyncConfig) {
     listener: config.offlineQueueListener,
     networkStatus: config.networkStatus as NetworkStatus
   });
-  const localDirectiveFilterLink = new LocalDirectiveFilterLink();
-  let offlineLinks = ApolloLink.from([offlineLink, localDirectiveFilterLink]);
-  offlineLinks = ApolloLink.split((op: Operation) => isMutation(op) && !isOnlineOnly(op), offlineLinks);
```
wtrocki: I kinda see what this stuff was doing. 
First it was making decision how queue was used (only for mutations and only when Operation is not online only).
We kinda missing that now.

StephenCoady: @wtrocki yes but in this scenario the mutation was still being added to the queue even when it was onlineOnly? the logic is off. 

now i do the check inside the offline link and don't enqueue. am i missing something or was that not the whole point of onlineOnly, that it was never queued and always just fired straight away?

wtrocki: Yes. This was because split was implemented incorrectly: 
https://www.apollographql.com/docs/link/composition.html#directional

packages/sync/integration_test/test/offline.test.js
```diff
@@ -24,7 +24,7 @@ const newClient = async (clientOptions = {}) => {
   return await createClient(config);
 };
 
-describe('AeroGear Apollo GraphQL Voyager client', function() {
+describe('AeroGear Apollo GraphQL Voyager client', function () {
```
wtrocki: Formatter issue?

StephenCoady: I think my beautify used the linter rules for the project as a guideline so I committed it.

PR #249, opened 2019-02-22
fix: fix broken conflict resolver check

StephenCoady:
small fix to add a default to the conflictResolutionStrategy Interface. This means that if the user provides a resolution function for a specific mutation then they do not need to name all as the default can also be defined.

### DIFF
--- a/packages/sync/src/config/SyncConfig.ts
+++ b/packages/sync/src/config/SyncConfig.ts
@@ -6,7 +6,7 @@ import { WebNetworkStatus } from "../offline";
 import { CordovaNetworkStatus } from "../offline";
 import { diffMergeClientWins } from "../conflicts/strategies";
 import { VersionedNextState } from "../conflicts/VersionedNextState";
-import { ConflictResolutionData } from "../conflicts/ConflictResolutionData";
+import { ConflictResolutionStrategies } from "../conflicts";
 
 declare var window: any;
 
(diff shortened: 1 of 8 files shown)


packages/sync/src/config/SyncConfig.ts
```diff
@@ -36,6 +37,13 @@ export class SyncConfig implements DataSyncConfig {
    * Method used to join user configuration with defaults
    */
   public merge(clientOptions?: DataSyncConfig): DataSyncConfig {
+    if (clientOptions && clientOptions.conflictStrategy) {
```
wtrocki: Not sure if the merge is the right place for doing this

packages/sync/src/conflicts/ConflictResolutionStrategy.ts
```diff
@@ -19,6 +19,6 @@ export type ConflictResolutionStrategy =
  * @param client - client data
  */
 export interface ConflictResolutionStrategies {
- [operationName: string]: (server: ConflictResolutionData,
-                           client: ConflictResolutionData) => ConflictResolutionData;
+ [operationName: string]: (server: ConflictResolutionData, client: ConflictResolutionData) => ConflictResolutionData;
+  default: (server: ConflictResolutionData, client: ConflictResolutionData) => ConflictResolutionData;
```
wtrocki: I'm not sure but it feels that this interface will overlap itself. 
[operationName: string] will match everything including "default" key. 
I would add a separate key for this just to avoid problems.

packages/sync/src/config/SyncConfig.ts
StephenCoady: sure, I can do that. I think the logic is ok because if conflictStrategy is not passed in the clientOptions object it is merged with the one set here: https://github.com/aerogear/aerogear-js-sdk/blob/366ce6d636398b5e17e53423db9134c4bb5029fa/packages/sync/src/config/SyncConfig.ts#L24

wtrocki: ok. So let's use this `conflictStrategy` inside here instead of defining a separate one. 
It will be a single place to make a change.

wtrocki:
This looks fantastic! going to verify after test fix.

PR #250, opened 2019-02-22
fix: Bump version of error link

wtrocki:
## Motivation 

Fix random crashes of the client when websockets are initiated too fast.
Errors that happening breaking entire client.

screen shot 2019-02-22 at 8 27 45 pm

The solution is to use the latest version of the error package that has this fixed:
https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-error/src/index.ts#L67-L70

### DIFF
--- a/packages/sync/package.json
+++ b/packages/sync/package.json
@@ -51,7 +51,7 @@
     "apollo-client": "2.4.5",
     "apollo-link": "1.2.3",
     "apollo-link-context": "^1.0.12",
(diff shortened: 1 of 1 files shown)


wtrocki:
Generally, we need to invest in updates of our packages. I have setup renovate in #248 but this will require a significant amount of time now to manually update typescript and deal with some compilation issues. 
I have that planned once things will settle.

PR #252, opened 2019-02-25
fix: minor dev dependency updates

darahayes:
##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


darahayes:
Looks like we're hitting this issue here also: https://github.com/highlightjs/highlight.js/issues/1984

PR #265, opened 2019-02-27
AEROGEAR-8560 check compile success during release validity

aliok:
Similar with https://github.com/aerogear/voyager-server/pull/126

@wtrocki In Voyager server if there's a compilation error, dist folder is empty. But not in JS sdk.
I couldn't figure out why.

Anyway, this is not the perfect check, but it would be thought as a smoke check only.


wtrocki:
@aliok How about we could check if `ls` on dist folder is not empty?

aliok:
@wtrocki there are JS files produced regardless the compilation error

PR #269, opened 2019-02-28
fix: implement sequential ordering of offline queue items

StephenCoady:
this is a small change to implement sequential ordering of the offline queue. this is needed to stop client data conflicting with itself.

### DIFF
--- a/packages/sync/src/offline/OfflineQueue.ts
+++ b/packages/sync/src/offline/OfflineQueue.ts
@@ -4,13 +4,14 @@ import { OfflineQueueListener } from "./OfflineQueueListener";
 import { OperationQueue, OperationQueueChangeHandler } from "./OperationQueue";
 import { isClientGeneratedId } from "../cache/createOptimisticResponse";
 import { squashOperations } from "./squashOperations";
-import { isSquashable, isMutation } from "../utils/helpers";
+import { ObjectState } from "../conflicts/ObjectState";
 
 export interface OfflineQueueOptions {
   storage?: PersistentStore<PersistedData>;
   storageKey?: string;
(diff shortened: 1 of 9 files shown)


wtrocki:
TL;DR:

Like PR as it evaluates on the best parts of the existing 
- We can get version thing abstracted
- We can lock processing to single operation as both ID's and versions are not updated now since all requests are sent at the same time 
(cc @jhellar Yeah I Know)
 

Open Questions:
- Should we remove squashing option?
- Should we update conflictInfo only when squashing is disabled

packages/sync/src/offline/OfflineQueue.ts
```diff
@@ -113,17 +116,18 @@ export class OfflineQueue extends OperationQueue {
...
-
       for (const { operation: op } of this.queue) {
         if (op.variables.id === entry.operation.variables.id && op.operationName === entry.operation.operationName) {
-          if (op.variables.version === entry.operation.variables.version) {
-            op.variables.version = newVersion;
+          const opVersion = this.state.currentState(op.variables);
+          const prevOpVersion = this.state.currentState(entry.operation.variables);
+          if (opVersion === prevOpVersion) {
+            const key = this.state.getKey();
+            op.variables[key] = this.state.currentState(this.state.nextState(op.variables));
```
wtrocki: We can avoid getKey method by simply providing method that will already know key. Literally pulling this logic into VersionState impl.

wtrocki:
Will be cool to get some better commit messages if possible.

StephenCoady:
@wtrocki this is ready for review if you have the time, thanks!

PR #270, opened 2019-02-28
Enable Retry by default

danielpassos:
Related issue: [AEROGEAR-8334](https://issues.jboss.org/browse/AEROGEAR-8334)


wtrocki:
We have disabled that by default because it was blocking responses and not everyone would like this. Is that no longer blocking?

wtrocki:
Closing as we removing retry logic.

PR #271, opened 2019-02-28
fix: Downgrade dependency for apollo client

wtrocki:
2.5 introducing breaking changes and after numerous of tests we found couple issues that will not be addressed till release.

# Verification

1. Link to showcase
2. Try offline edit.
3. Restore internet connection


wtrocki:
2.4.5 is not latest but it was proven and tested.

danielpassos:
@wtrocki what is the best way to test it?

wtrocki:
No longer needed. I have fixed problems with underlying integration and things seems to be working fine.

PR #279, opened 2019-03-05
Process queue only when offline

wtrocki:
## Motivation

The queue should be processed when offline only.
Mutations should be marked as offline to enqueue them from offline mutation handler. Posting for early feedback. Please shout as soon as you see something wrong

## Progress

- [x] Forward online request and let them fail fast
- [x] Enable retry link for offline handler
- [x] Do not dequeue when retrylink returned error
- [x] Provide safe defaults for retry link
- [x] Connect mutation listeners to retry link
- [x] Remove config for retry link (this should be internal

## Testing

I have performed testing for all possible cases but feel free to chime in with testing.

### Cases

- Offline changes restored when server is down
- Offline changes ending up with graphql error
- Offline changes are successful. 

## Not included (separate PR) 

- [ ] Pass optimistic response when saving offline change (this will resolve the problem with disappearing optimistic responses)

### DIFF
--- a/packages/sync/src/offline/retry/RetriableOperation.ts
+++ b/packages/sync/src/offline/retry/RetriableOperation.ts
@@ -47,7 +47,7 @@ export class RetriableOperation extends OperationQueueEntry {
             setTimeout(resolve, this.timeout);
           });
 
-          this.timeout *= 2;
+          this.timeout *= 3;
           attempts++;
 
           retry = shouldRetry(attempts, this.operation, error);
--- a/packages/sync/src/offline/retry/RetriableOperation.ts
+++ b/packages/sync/src/offline/retry/RetriableOperation.ts
@@ -1,11 +1,9 @@
-import { isNetworkError } from "../utils/helpers";
-import { OperationQueueEntry } from "./OperationQueueEntry";
-import { Operation } from "apollo-link";
+import { isNetworkError } from "../../utils/helpers";
+import { OperationQueueEntry } from "../OperationQueueEntry";
+import { ShouldRetryFn } from "./ShouldRetry";
 
 // Initial timeout for first retry
 const INITIAL_TIMEOUT = 1000;
-
-export type shouldRetryFn = (count: number, operation: Operation, error: any) => boolean;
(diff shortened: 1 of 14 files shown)


wtrocki:
@jhellar I finished first round of improvements. At this point I think that all works quite nice.

We do not touch non offline changes.
We adding flag for offline one so they properly enqueued by offline restore handler
We using retry link only internally to retry the offline changes only.
We do not dequeue failed changes (I know that this means that they will stay there for a while, but it is safer than data los)
We also notify users about the whole process success and failures.

Feel free to post some feedback while'm going to work on writing and fixing tests. I think that we could do second part (satisfying then/catch for offline) in separate PR.

wtrocki:
@darahayes I needed your head around this as well as @StephenCoady is on holidays

jhellar:
The failing test case is checking that when device is restarted, and offline mutations are sent to server, id updating works - e.g. when there was create and then update mutation on the same item. I think that the test case is failing because of this [1]. The information that the mutation is offline is in the local storage (in extensions), but it is not used in `OfflineRestoreHandler`, so when the mutation comes to OfflineLink it will not have the `markedOffline` flag, and it will be forwarded to next link immediately.

[1] https://github.com/aerogear/aerogear-js-sdk/blob/f6f82eb962d6dddea41568c803a35e66fb17a750/packages/sync/src/offline/OfflineRestoreHandler.ts#L50-L53

wtrocki:
I have tested all cases:

- network error
- graphql errors
- successful calls

Everything works as expected.

packages/sync/src/offline/retry/RetriableOperation.ts
```diff
@@ -49,7 +47,7 @@ export class RetriableOperation extends OperationQueueEntry {
             setTimeout(resolve, this.timeout);
           });
 
-          this.timeout *= 2;
+          this.timeout *= 3;
```
darahayes: minor comment: maybe this should be a constant variable at the top of the file like `TIMEOUT_MULTIPLIER`

PR #286, opened 2019-03-08
Update OfflineLink.ts

wtrocki:
It looks like dequeue was calling forward operation when online. 
From my way of understanding if feels that this is not needed as dequeue is called only when operation was forwarded so server. 
Any opinions?


jhellar:
The offline queue does not forward mutations with client generated id. The idea was that whenever operation is dequeued (response from server), offlineLink should check if some of the mutations in offlineLink are ready now (if some operation with client id was updated with server id). And forward all those mutations.

wtrocki:
@jhellar I have played with that on Friday and yes that is the case. There is no simple fix here. For the long run, I would like to simplify entire flow and drop `toBeForwarded` method together with forward on dequeue.

PR #292, opened 2019-03-12
Add ability to pass update functions to the restore handler

wtrocki:
The first iteration heavily inspired by https://medium.com/twostoryrobot/a-recipe-for-offline-support-in-react-apollo-571ad7e6f7f4

I haven't resolved "chicken or egg problem" related to the fact that the client is usually created first and then update functions follow. This will mean that our update functions need to be pulled into a separate class (later we can replace them with the helper that generates both optimistic responses and also update function. That will greatly improve the user experience.  

EDIT: It is actually easy to describe that now and it has very little impact on users. I will need to create follow up ticket to document this pattern.

## Tasks
- [x] Refactor previous implementation
- [x] Integrate with a showcase
- [x] Add case to integration tests


StephenCoady:
@wtrocki this all makes sense and looks good to me. will wait until i try it out with the showcase

wtrocki:
Showcase is ready: https://github.com/aerogear/ionic-showcase/pull/104
Going to do development release to make build pass

packages/sync/src/offline/OfflineRestoreHandler.ts
```diff
@@ -48,14 +52,21 @@ export class OfflineRestoreHandler {
...
+      const mutationName = getMutationName(item.operation.query);
+      let updateFunction;
+      if (this.mutationCacheUpdates && mutationName) {
+        updateFunction = this.mutationCacheUpdates[mutationName];
+      }
+      const mutationOptions = {
         variables: item.operation.variables,
         mutation: item.operation.query,
         // Restore optimistic response from operation in order to see it
         optimisticResponse,
```
wei-lee: this is just question from me. Given that we have the `optimisticResponse` here, can we actually use it to update the UI?

packages/sync/src/utils/helpers.ts
```diff
@@ -35,3 +36,9 @@ export const isOnlineOnly = (op: Operation) => {
 export const isNetworkError = (error: any) => {
   return !error.result;
 };
+
+export const getMutationName = (mutation: DocumentNode) => {
+  const definition = mutation.definitions.find(def => def.kind === "OperationDefinition");
+  const operationDefinition = definition && definition as OperationDefinitionNode;
+  return operationDefinition && operationDefinition.name && operationDefinition.name.value;
```
wei-lee: is it possible if the operation doesn't have a name?

packages/sync/src/offline/OfflineRestoreHandler.ts
wtrocki: Good question. The optimistic response is saved automatically and does not need any update method. Cache layer will also apply some heuristics to figure out affected queries (this will happen objects that are currently observed for current queries).
This is just an plain `Type` object response. We do not know what query this mutation will affect if there is a need for subscription deduplication etc.

UI is being updated thru cache subscriptions in showcase app - we need to make sure that this lands in the cache as it will be the only way to support many different query patterns on framework level + provide the ability to do edits on that objects as well. Skipping cache will cause many side effects and will require us to provide some custom handling. 

Technically we can use object enqueue listener instead of the update function, but we still need to map actual response from server anyway so it will mean doing twice the job. It is really good idea to consider:
Provide our own form of update functions that skip cache and simply deliver response to UI using some new customized observable stream. On startup we can rebuild entire stream and deliver those updates again however by this we going to need to largely reimplement optimistic response merging functionality from InMemory cache. That is not something I could get in short timeframe.

packages/sync/src/utils/helpers.ts
wtrocki: Acording to spec - Not for mutation. Operation is generic and handles queries, subscriptions and mutations. Queries can exist without explicity name but this is fairly uncommon thing anyway.

wtrocki:
Going to create separate ticket to document that new option.

PR #288, opened 2019-03-11
AEROGEAR-8664

austincunningham:
# Description
https://issues.jboss.org/browse/AEROGEAR-8772
Create a class in the security-services that uses the aerogear-js-sdk init for enabling and disabling

Implementation is similar to the Auth package
### auth
https://github.com/aerogear/aerogear-js-sdk/blob/964a45b0a851720410b68958774ed0f97e7d6785/packages/auth/src/Auth.ts#L17

### push
https://github.com/aerogear/aerogear-js-sdk/blob/964a45b0a851720410b68958774ed0f97e7d6785/packages/push/src/PushRegistration.ts#L23

### sync
https://github.com/aerogear/aerogear-js-sdk/blob/964a45b0a851720410b68958774ed0f97e7d6785/packages/sync/src/config/SyncConfig.ts#L26

No Verification steps as this is a constructor 

##### Checklist

- [x] `npm test` passes

PR #293, opened 2019-03-12
Revert "AEROGEAR-8664"

wtrocki:
Reverts aerogear/aerogear-js-sdk#288

@austincunningham Would it be possible to keep this work on feature branch until our release? We need one week to do a major release and then all the work will be unblocked. The rule is that every change that lands to master should be functional and testable.


wtrocki:
Ignore me :). Just seen that this was merged to different branch!

PR #298, opened 2019-03-13
Aerogear 8791

StephenCoady:
## Motivation

Simplify observables flow.

## Tasks

- [x] Offline restore handler should not use for each
and a stop wiping the data. We need to check if changes were offline
before persisting and skip it.

- [x] Refactor OperationQueueEntry and replace it with the simple
OfflineQueueEntry that will contain only operation and optimistic response.

- [x] Disable and Retry link for offline changes instead of own.

- [x] Remove toBeForwarded 

- [x] Merge OfflineQueue with OperationQueue

- [x] Add open/closed actions

- [x] Figure out best pattern for sequential processing of observables. 

- [x] Connect proper observables flow

IMG_20190313_165922

- [x]  Fix unit tests and integration tests


wtrocki:
This should work now apart from OfflineRestoreHandler (on restart). 
It looks like this never worked universally and it was only because of luck/timing of processing was just coincidentally making this work

wtrocki:
Outstanding tasks:

- [ ] Test that on mobile (android)
- [ ] Create separate to add ApolloLinkRetry for offline changes (we removed our own homebrew implementation here.
- [ ] Throw a party with lots of wine and food

wtrocki:
Testing that stuff now in browser and android.

packages/sync/src/createClient.ts
```diff
@@ -21,20 +22,30 @@ export const createClient = async (userConfig?: DataSyncConfig): Promise<Voyager
   const clientConfig = extractConfig(userConfig);
   const { cache } = await buildCachePersistence(clientConfig);
 
-  const link = await defaultLink(clientConfig);
+  const offlineLink = await createOfflineLink(clientConfig);
```
wtrocki: This gives a lot of flexibility now however we are no longer have all logic in links. 
restoreOfflineOperations functions are external and require access to the link.

packages/sync/src/offline/OfflineQueue.ts
```diff
@@ -23,79 +27,91 @@ export interface OfflineQueueOptions {
...
-    this.storage = storage;
-    this.storageKey = storageKey;
-    this.listener = listener;
-    this.state = conflictStateProvider;
+  public enqueue(operation: Operation, forward: NextLink) {
+    const operationEntry = new OperationQueueEntry(operation, forward);
+    this.enqueueEntry(operationEntry);
+    return new Observable((observer) => {
+      operationEntry.observer = observer;
+      return () => ({ isOffline: true });
```
wtrocki: isOffline: true is confusing as it is not used for anything at the moment. 
This function is getting called when client unsubscribes which in context of the links is going to be called only when client cancels request (like someone called destroy or something).
With that in mind I think we should not support canceling those for the moment and leave TODO like in http link: 
https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-http/src/httpLink.ts#L187-L191

@StephenCoady WDYT?

packages/sync/src/offline/OfflineQueue.ts
```diff
@@ -23,79 +27,91 @@ export interface OfflineQueueOptions {
...
-   */
-  public toBeForwarded() {
-    return this.queue.filter(op => !op.subscription && !op.hasClientId());
+  public async forwardOperations() {
+    for (const op of this.queue) {
+      // FIXME block operations till result is back (completed)
+      await new Promise((resolve, reject) => {
+        op.forward(op.operation).subscribe({
+          next: (result: FetchResult) => {
+            this.queue = this.queue.filter(e => e !== op);
```
wtrocki: This is hard to read. I would love to have next and error handled by separate methods for better readability.

packages/sync/src/offline/OfflineQueue.ts
```diff
@@ -23,79 +27,91 @@ export interface OfflineQueueOptions {
...
+
+            if (this.queue.length === 0 && this.listener && this.listener.queueCleared) {
+              this.listener.queueCleared();
+            }
+            if (op.observer) {
+              op.observer.next(result);
+            }
+          },
+          error: (error: any) => {
+            // TODO remove result from operation and handle that directly.
```
wtrocki: Not actual TODO. We have done that already :D

packages/sync/src/offline/OfflineQueue.ts
```diff
@@ -23,79 +27,91 @@ export interface OfflineQueueOptions {
...
+            if (error) {
+              if (this.listener && this.listener.onOperationFailure) {
+                this.listener.onOperationFailure(op.operation, undefined, op.networkError);
+              }
+              return;
+            }
+            if (op.observer) {
+              op.observer.error(error);
+            }
+            // TODO - do we want to handle try catch here?
```
wtrocki: Did I wrote that. If yes then let's remove it as I do not know what `handle try catch` can mean in this context. There is no error thrown.

packages/sync/src/offline/OfflineQueue.ts
wtrocki:
```suggestion
      return () => { 
        // TODO support for canceling offline requests.
        return; 
      }
```

PR #301, opened 2019-03-14
Extend integration tests for data sync - realtime updates

jhellar:
### Description

https://issues.jboss.org/browse/AEROGEAR-8802


packages/sync/integration_test/test/subscriptions.test.js
```diff
@@ -0,0 +1,126 @@
...
+    title: 'custom'
+  }
+};
+
+const newClient = async (clientOptions = {}) => {
+  const config = {
+    httpUrl: "http://localhost:4000/graphql",
+    wsUrl: "ws://localhost:4000/graphql",
+    ...clientOptions,
+    conflictStrategy: {
```
wtrocki: Are we using any conflict stuff for subscriptions?

jhellar: no, I've copy pasted the config, it's not needed, will remove it

PR #303, opened 2019-03-14
fix: fix build that was failing on linux(AEROGEAR-8800)

austincunningham:
### Description
build failing on linux see the following gist for error https://gist.github.com/austincunningham/856e542ea421ec86c2ce3a54ac6fcf13  

adding dom to the tsconfig fixes the issue
##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,7 +10,8 @@
     "lib": [
       "es6",
       "es2015",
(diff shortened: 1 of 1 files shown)


austincunningham:
@wtrocki as requested

PR #308, opened 2019-03-21
AEROGEAR-8773 get device data

austincunningham:
### Description

Merging to feature branch `app-security`
# What
PR Gets device data and app data for the security services app-service module
# How
Uses MetricsBuilder

# Why
Needed for the Mobile security service Go server app-security integration
https://issues.jboss.org/browse/AEROGEAR-8773

# Verification
## Prerequisites 
- node 10
- npm 6
- install ionic https://ionicframework.com/docs/installation/cli
- install android sdk 26 http://www.androiddocs.com/sdk/installing/adding-packages.html

## Steps
- clone this repo
- checkout branch AEROGEAR-8773
- install the aerogear-js-sdk
```bash
npm i && npm run bootstrap && npm run build
```

- clone https://github.com/austincunningham/ionic-showcase.git
- checkout branch AEROGEAR-8773
- change the package.json to point at your local clone of aerogear-js-sdk e.g.

image

> Note you can try and use npm link but I didn't have much joy with it 
- install ionic-showcase
```
npm i
```
- build the android apk
```
ionic cordova build android
```
- copy the apk to your android device emulation
- open the app on the emulator
- follow these steps to attach your chrome developer console to your emulation https://developers.google.com/web/tools/chrome-devtools/remote-debugging/

- the console you should see the Client Data and the metrics payload to verify this

remotedebug

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works

### DIFF
--- a/packages/security/src/appSecurity/AppSecurity.ts
+++ b/packages/security/src/appSecurity/AppSecurity.ts
@@ -29,7 +29,7 @@ export class AppSecurity {
       metricsPayload[metric.identifier] = await metric.collect();
     }
     metricsPayload.clientID = metricsBuilder.getClientId();
-    return  {metricsPayload};
+    return metricsPayload;
   }
 
   // TODO call init endpoint on go mobile security service AEROGEAR-8774
--- a/packages/security/src/appSecurity/AppSecurity.ts
+++ b/packages/security/src/appSecurity/AppSecurity.ts
@@ -15,10 +15,10 @@ export class AppSecurity {
     } else {
       console.warn("Security configuration is missing. The Security module will not work properly.");
     }
-    console.info("Url: ", this.internalConfig.url);
   }
 
   // Get device info AEROGEAR-8773
+  // TODO move this functionality to the core
   public getClientData = async (): Promise<any> => {
     const metricsBuilder: MetricsBuilder = new MetricsBuilder();
     const metricsPayload: {
--- a/packages/security/src/appSecurity/AppSecurity.ts
+++ b/packages/security/src/appSecurity/AppSecurity.ts
@@ -28,6 +28,7 @@ export class AppSecurity {
     for (const metric of metrics) {
       metricsPayload[metric.identifier] = await metric.collect();
     }
+    metricsPayload.clientID = metricsBuilder.getClientId();
     return  {metricsPayload};
   }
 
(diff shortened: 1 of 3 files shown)


packages/security/src/appSecurity/AppSecurity.ts
```diff
@@ -15,8 +15,21 @@ export class AppSecurity {
     } else {
       console.warn("Security configuration is missing. The Security module will not work properly.");
     }
+    console.info("Url: ", this.internalConfig.url);
```
wtrocki: I would remove that or add more info to the header and debug type.

packages/security/src/appSecurity/AppSecurity.ts
```diff
@@ -15,8 +15,21 @@ export class AppSecurity {
...
+  }
+
+  // Get device info AEROGEAR-8773
+  public getClientData = async (): Promise<any> => {
+    const metricsBuilder: MetricsBuilder = new MetricsBuilder();
+    const metricsPayload: {
+      [key: string]: any;
+    } = {};
+    const metrics = metricsBuilder.buildDefaultMetrics();
+    for (const metric of metrics) {
```
wtrocki: We are leaking this logic from metrics SDK.
We could add this as a separate method to the builder and simply reuse core method: https://github.com/aerogear/aerogear-js-sdk/blob/master/packages/core/src/metrics/MetricsService.ts#L101-L104

I think the requirement here is to build payload but not send it to the server so it makes sense (IMHO) to just extract and reuse the same function.

wtrocki:
Another idea is to use builder as it is but with different `MetricsPublisher` that simply publishes to your own server. With that in mind, all core implementations will be sufficient.

packages/security/src/appSecurity/AppSecurity.ts
austincunningham: :+1:

packages/security/src/appSecurity/AppSecurity.ts
austincunningham: While I agree that it would be better to have this functionality in metrics in the core. While we are developing on a feature branch it makes more sense to keep the code bases separate to avoid having to constantly rebase and fix conflicts on the feature branch. I will spin up a jira to move this functionally to the core once we have merged with master @wtrocki what do you think ?

wtrocki: `code bases separate to avoid having to constantly rebase and fix conflicts on the feature branch` 

FYI There is no active development on js-sdk at the moment and there will be no active dev in the next 3 months. This functionality is just 3 lines that exist in the code. It is faster to move that than create jira.

craicoverflow:
Failing on `npm install` for the ionic-showcase app:

```sh
npm ERR! Could not install from "@aerogear/cordova-plugin-aerogear-metrics@~2.1.0-rc1" as it does not contain a package.json file.
```

wtrocki:
As changes are hard to test now it is best to just continue working on this and then we going to review and test the actual feature branch.

PR #310, opened 2019-03-25
feat: added retryLink for offline queued mutation failures

StephenCoady:
### Description

added the apollo retry link to retry mutations which have failed and were previously queued offline.

### Verification
1. Go offline
1. Create a couple of mutations
1. Stop the server
1. Go back online
1. Mutations should be attempted to be replayed after 1 second and should be retried 5 times. If they are unsuccessful they should remain in the queue. If they are successful they should be removed from the offline queue. 

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/sync/src/config/SyncConfig.ts
+++ b/packages/sync/src/config/SyncConfig.ts
@@ -45,6 +45,17 @@ export class SyncConfig implements DataSyncConfig {
   public conflictStrategy: ConflictResolutionStrategies;
   public conflictStateProvider = new VersionedState();
 
+  public retryOptions = {
+    delay: {
+      initial: 1000,
+      max: Infinity,
+      jitter: true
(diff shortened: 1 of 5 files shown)


packages/sync/src/config/SyncConfig.ts
```diff
@@ -45,6 +45,17 @@ export class SyncConfig implements DataSyncConfig {
   public conflictStrategy: ConflictResolutionStrategies;
   public conflictStateProvider = new VersionedState();
 
+  public retryOptions = {
+    delay: {
+      initial: 1000,
+      max: Infinity,
```
wtrocki: [Minor] Extract that to separate file so it will not take so much space here.

PR #312, opened 2019-03-27
fix: Update showcase application link

wtrocki:
Minor fix for documentation pointing to recent showcase application

### DIFF
--- a/docs/contrib/contributing-guide.adoc
+++ b/docs/contrib/contributing-guide.adoc
@@ -52,7 +52,7 @@ Please refer to the https://github.com/aerogear/aerogear-js-sdk/blob/master/docs
 
 Testing can be done using the showcase application:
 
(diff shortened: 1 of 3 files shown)


docs/contrib/contributing-guide.adoc
```diff
@@ -52,22 +52,11 @@ Please refer to the https://github.com/aerogear/aerogear-js-sdk/blob/master/docs
...
 npm run build
 npm run link
-
-If you wish, you may then also use the provided script cloneExamples.sh to clone the showcase application into the SDK's
-root folder.
-
-Once you have the showcase application cloned you can then run the following from within its root folder:
-
-[source, bash]
-npm install
```
wtrocki: This is now much more complex than before since AeroGear Server packages also need to be linked. Developers will link only packages they are interested in so using this script makes less sense.

PR #316, opened 2019-04-04
feat: swapped storage to localForage

StephenCoady:
### Description
This PR removes localStorage as the sdks client side storage driver. It replaces it with localForage which is much more performant and uses IndexedDB under the hood.

### Verification
To verify that this PR works, clone the showcase application and npm link this PR into it.

* Perform some actions while online, these actions are using the cache and storing it in the local db and you should be able to see them in the console like so:

Screenshot 2019-04-04 16 38 27

* go offline and perform some actions, they should appear in the db in the offline store. note you may need to refresh the view to see them.

Screenshot 2019-04-04 16 40 58

* Next, configure the storageOptions param in the config on the client with something like:
```
storageOptions = {
     name: "prTestName",
     storeName: "prStoreTestName" 
}
```
now when you perform the same actions above the db and store names should be different.

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


packages/sync/src/config/DataSyncConfig.ts
```diff
@@ -118,3 +124,8 @@ export interface DataSyncConfig {
    */
   retryOptions?: RetryLink.Options;
 }
+
+export interface StorageOptions {
```
wtrocki: We can avoid having another option. Simply we can allow people to simply pass localforage instance that is suited to their needs. By default we can use some internal localforage instance with sensible defaults.

packages/sync/src/config/SyncConfig.ts
```diff
@@ -39,7 +38,11 @@ export class SyncConfig implements DataSyncConfig {
     }
   }
 
-  public storage?: PersistentStore<PersistedData>;
+  public storage?: LocalForage;
+  public storageOptions = {
+    name: "applicationDB",
```
wtrocki: Having it this way gives no benefit over local storage. 
Do you think we can try different `driver` option?

See: https://github.com/localForage/localForage

packages/sync/src/offline/OfflineRestoreHandler.ts
```diff
@@ -18,13 +17,13 @@ export const logger = debug.default(MUTATION_QUEUE_LOGGER);
...
   private apolloClient: ApolloClient<NormalizedCacheObject>;
-  private storage: PersistentStore<PersistedData>;
+  private storage: LocalForage;
   private readonly storageKey: string;
   private mutationCacheUpdates?: CacheUpdates;
 
   constructor(apolloClient: ApolloClient<NormalizedCacheObject>, clientConfig: DataSyncConfig) {
     this.apolloClient = apolloClient;
-    this.storage = clientConfig.storage as PersistentStore<PersistedData>;
+    this.storage = clientConfig.storage as LocalForage;
```
wtrocki: Any reason for using LocalForage? Using this explicitly can affect user interface - they will be able to use LocalForage only when we use some methods that aren't in the PersistentStore interface.

packages/sync/src/config/SyncConfig.ts
StephenCoady: @wtrocki you can see the default order here: https://github.com/localForage/localForage/blob/master/src/localforage.js#L23-L27

it uses IndexedDB first

packages/sync/src/offline/OfflineRestoreHandler.ts
StephenCoady: OK maybe I got it wrong but to me it looked like our PersistentStore interface only had 3 methods and all of these are covered properly by LocalForage. Was I wrong? Or do you mean what if they want to use something else as storage?

StephenCoady:
@wtrocki do you think you could try this out for me locally when you get a chance? I am a bit stumped as the best way to solve it.

At the moment I am saving each operation as the id of the object. I see the problem with this now based on what you said yesterday about using objects instead. If I store based on ids then multiple actions on the same object will be lost (i.e. create then update will only result in the update being saved.) 

I know this PR is rough but I can clean up the config stuff later. If I can get a good storage model I can fix the rest of it later.

PR #325, opened 2019-04-11
feat: swapped out custom interface for Localforage

StephenCoady:
### Description
This PR adds the ability to use async storage as default in the sdk. 

It also splits out the offline mutation store into individual mutations. This reduces the need to stringify all of these mutations each time one is added or read.

Verification:

1. Use the showcase application and default config.
1. Clone this branch, build it and npm link it to the showcase application by: `npm link @aerogear/voyager-client`
1. Run the showcase application with `npm run start`
1. Create some items and everything should work fine
1. Go offline
1. Create some more items
1. Check async storage in your browser and you should see multiple entries depending on how many you created
1. Go online
1. Items should be replayed and synced to the server
1. Go offline
1. Create some items
1. Go online and restart the application
1. The previous offline items should be replayed and synced to the server


StephenCoady:
Re-opening this after a conversation with @wei-lee. All that is left to add here is the logic to allow the developer to choose which driver they want to use within localForage without needing a completely new one.

wtrocki:
Leaving for @wei-lee to make the final call here. 
From what I see this works exactly like before but with indexedDb used as the base.

packages/sync/src/config/DataSyncConfig.ts
```diff
@@ -33,7 +32,7 @@ export interface DataSyncConfig {
    *
    * The storage you want your client to use (Uses window.localStorage by default)
    */
-  storage?: PersistentStore<PersistedData>;
+  storage?: LocalForage;
```
wei-lee: do we have to change the type here? It should continue to be the `PersistentStore<PersistedData>` interface, and LocalForage is an implementation of this interface?

packages/sync/src/config/SyncConfig.ts
```diff
@@ -60,9 +61,8 @@ export class SyncConfig implements DataSyncConfig {
   private readonly clientConfig: DataSyncConfig;
 
   constructor(clientOptions?: DataSyncConfig) {
-    if (window) {
-      this.storage = window.localStorage;
-    }
+    LocalForage.config({});
```
wei-lee: the option here should be configurable.

packages/sync/src/createClient.ts
```diff
@@ -64,7 +64,7 @@ async function buildCachePersistence(clientConfig: DataSyncConfig) {
   const cache = new InMemoryCache();
   await persistCache({
     cache,
-    storage: clientConfig.storage as PersistentStore<PersistedData>,
+    storage: clientConfig.storage as PersistentStorage<PersistedData<NormalizedCacheObject>>,
```
wei-lee: if the type is not changed, then you don't need the cast here.

packages/sync/src/config/SyncConfig.ts
```diff
@@ -95,4 +101,23 @@ export class SyncConfig implements DataSyncConfig {
     return Object.assign(this, clientOptions);
   }
 
+  /**
+   * A method to setup the storage. Defaults to IndexedDB wrapped by LocalForage if none is passed
+   * @param clientOptions the options passed from the constructor containing the storage options
+   */
+  private setStorage(clientOptions?: DataSyncConfig) {
+    LocalForage.config(this.storageOptions);
```
wtrocki: This is not needed.

StephenCoady: You have to call config before you call createInstance or else you risk getItem, setItem etc not working correctly.

StephenCoady:
ping @wtrocki if you have a chance to clone this locally and play with it. its somewhat rough but it is working for me now. let me know what you think.

wtrocki:
👀

packages/sync/src/PersistentStore.ts
```diff
@@ -3,9 +3,9 @@
  * Interface for underlying storage solutions
  */
 export interface PersistentStore<T> {
-  getItem: (key: string) => Promise<T> | T;
-  setItem: (key: string, data: T) => Promise<void> | void;
-  removeItem: (key: string) => Promise<void> | void;
+  getItem: (key: string) => Promise<T>;
+  setItem: (key: string, data: T) => any;
+  removeItem: (key: string) => Promise<void>;
```
wtrocki: It doesn't look for me that these changes were essential to get this work finished. Is that just an experiment or this is required?

packages/sync/src/config/DataSyncConfig.ts
```diff
@@ -35,6 +36,20 @@ export interface DataSyncConfig {
...
+   * The storage the offline client will use to persist offline mutations
+   */
+  offlineStorage?: PersistentStore<OfflineItem>;
+
+  /**
+   * [Modifier]
+   *
+   * The storage options you want the default driver to use. Ignored if `storage` is passed in config.
+   */
+  storageOptions?: StorageOptions;
```
wtrocki: Feels really strange to have localforage specific options (undocumented here) together with the ability to swap storage. 

No strong opinion but from my point of view it is important to figure out defaults and let people pass custom storage if they want it. 

There will be much better storage solutions for Cordova and React native so localforage will not always be the answer, but it could be sensible default for web and cordova for now.

packages/sync/src/offline/OfflineStore.ts
```diff
@@ -7,39 +7,51 @@ import { OperationQueueEntry, OfflineItem } from "./OperationQueueEntry";
...
-    if (stored) {
-      offlineData = JSON.parse(stored.toString());
-    } else {
-      offlineData = [];
+    const keys = await this.storage.getItem("offlineKeys");
+    const offlineItems: OfflineItem[] = [];
+    if (keys) {
+      this.arrayOfKeys = JSON.parse(keys.toString());
+      for (const key of this.arrayOfKeys) {
+        const item = await this.offlineStorage.getItem(key.toString());
```
wtrocki: This breaks `window.localStorage` as it returns string etc.
It will be best to check if returned object is string or object and then apply stringify if needed.

wtrocki: Same AsyncStorage from react: https://github.com/react-native-community/react-native-async-storage/blob/master/lib/AsyncStorage.js#L75

wtrocki:
Cloned this locally. I will check how this will behave with localStorage and AsyncStorage. Generally looks good. I would go with further simplification of the interface.  From the config perspective lets not make it overly complex as it will be hard to support. Imagine that we need to ask many different questions.
Since we have now ability for people to extend class we do not need separate config right?

packages/sync/src/PersistentStore.ts
StephenCoady: They are unfortunately. Our old interface didn't match Localforage exactly. There is a github issue somewhere, I'll try to find it

packages/sync/src/config/DataSyncConfig.ts
StephenCoady: OK, yeah good point. I will set it to configure defaults unless they provide the storage themselves.

packages/sync/src/PersistentStore.ts
wtrocki: It does not need to match local forage it needs to match the Apollo one. I will check what was failing in particular here and propose something. This one will break other providers as we made it less flexible rath

packages/sync/src/PersistentStore.ts
```diff
@@ -3,9 +3,9 @@
...
 }
 
 export type PersistedData = string | null | object;
```
wtrocki: @StephenCoady We need to aling Storage for data to be string and need to be stringified or object.

StephenCoady:
@wtrocki ready for final review if you have a chance.

wtrocki:
Closing in favor of #344

PR #326, opened 2019-04-11
feat: added storage wrapper to enable encryption of local data

StephenCoady:
### Description

image of encrypted data in localStorage:
Screenshot 2019-04-11 12 44 42

The flow of this is now:

1. Client provides a password
2. If the app starts and there is no masterKey wrapped in storage then this code generates one, wraps it, and stores it.
3. If the app starts and tere is a masterKey the app unwraps it from storage.
4. The app uses this key so app restarts will not affect the key

### Verification

* clone this branch locally
* clone this branch: https://github.com/StephenCoady/ionic-showcase/tree/add-encryption of the showcase application and link the sdk to there
* provide an `encryptionPassword` in the config of the app
* npm run start the showcase application and the storage should resemble that of the screen shot above.
* make some offline changes and they should also be encrypted
* return online and the changes should be decrypted and synced correctly

An example of they key to provide from the app is: 

and then pass this object under DataSyncConfig as encryptionKey
##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/sync/src/PersistentStore.ts
+++ b/packages/sync/src/PersistentStore.ts
@@ -1,4 +1,3 @@
-
 /**
  * Interface for underlying storage solutions
  */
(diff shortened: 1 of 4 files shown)


wtrocki:
This perfect! Really good. Before we merge I will check how this works on mobile in terms of perf and support

wtrocki:
Coverals failing. Maybe we can unit test that

packages/sync/package.json
```diff
@@ -55,6 +56,7 @@
     "apollo-link-retry": "2.2.13",
     "apollo-link-ws": "1.0.17",
     "apollo-upload-client": "10.0.0",
+    "crypto-js": "^3.1.9-1",
```
wtrocki: I think we can do better without doing extensive benchmarks:

https://www.npmjs.com/package/webcrypto

https://medium.com/@encryb/comparing-performance-of-javascript-cryptography-libraries-42fb138116f3

wei-lee:
We probably should consider using web crypto API instead of using the CryptoJS. There are a few benefits of use it:

1. It is a webstandard and looks like supported by many browsers.
2. We don't need to include another lib.
3. The APIs are async and it can improve performance.

packages/sync/src/config/SyncConfig.ts
```diff
@@ -77,6 +81,10 @@ export class SyncConfig implements DataSyncConfig {
     this.clientConfig = this.init(clientOptions);
   }
 
+  public async getStorage() {
+    this.storage = await new StorageWrapper(window.localStorage, "9E715A41D45BA00A3D410C23BE436DAA");
```
StephenCoady: I know, I've hardcoded the key for now. just makes it easier for testing and won't matter as this won't be merged

StephenCoady:
@wtrocki @wei-lee I have swapped this to use the web crypto api now.

If you guys get a chance could you take a look please. I know this won't be merged here but I am pushing here as it is easier to review from a PR. If you are both happy with this I will push it to a separate package.

@wei-lee I know we spoke about wrapping the key so that it persisted. I implemented this but the unwrapping was very tricky and I got a lot of errors. In the end I decided it wasn't worth the work to implement so I abandoned it. In its current form the user will need to pass a jwk and then everything should work fine. If you are interested in the wrapping code I wrote its here: https://gist.github.com/StephenCoady/f815bc00f8167c58e756412313998e64

packages/sync/package.json
wei-lee: @StephenCoady is this still required?

StephenCoady: @wei-lee no, I removed it in another commit I forgot to push sorry.

StephenCoady:
The tests are failing because I have left in a console.log.

If this is ok @wei-lee I will extract it to a cleaner module with a few files to make it easier to read and push somewhere. I will add some unit tests then as well.

If you want to try it out I have pushed to the showcase branch listed in the verification steps above.

packages/sync/src/PersistentStore.ts
```diff
@@ -9,3 +8,199 @@ export interface PersistentStore<T> {
...
+export class StorageWrapper implements PersistentStore<PersistedData> {
+  public storage: PersistentStore<PersistedData>;
+  private crypto = window.crypto;
+  private masterKey: CryptoKey | undefined;
+  private password: string;
+
+  constructor(storage: PersistentStore<PersistedData>, encryptionPassword: string) {
+    this.password = encryptionPassword;
+    this.storage = storage;
+    this.init();
```
wei-lee: not sure if it's a good idea to call `init` here and but not wait for it to complete. It could cause some issues depending on how it is invoked. I suggest move the `init` call into `getItem` and `setItem` method.

packages/sync/src/PersistentStore.ts
```diff
@@ -9,3 +8,199 @@ export interface PersistentStore<T> {
...
+   * @param key the key to use to retreive your value
+   */
+  public getItem(key: string) {
+    const retrieved = this.storage.getItem(key);
+    if (retrieved) {
+      if (this.masterKey) {
+        const cipherTextWithIv = retrieved.toString();
+        const cipherText = cipherTextWithIv.slice(cipherTextWithIv.indexOf(":") + 1);
+        const iv = this.strToArray(cipherTextWithIv.substr(0, cipherTextWithIv.indexOf(":")));
+        this.crypto.subtle.decrypt({ name: "AES-GCM", iv },
```
wei-lee: it might be better to move L33-36 into its own function to make the `getItem` function easier to read.

packages/sync/src/PersistentStore.ts
```diff
@@ -9,3 +8,199 @@ export interface PersistentStore<T> {
...
+    const iv = crypto.getRandomValues(new Uint8Array(12));
+    if (data) {
+      if (this.masterKey) {
+        this.crypto.subtle.encrypt({
+          name: "AES-GCM",
+          iv
+        },
+          this.masterKey,
+          this.strToArray(data.toString())
+        ).then(cipherText => {
```
wei-lee: same here, move L52-61 into its own function to make `setItem` easier to read.

packages/sync/src/PersistentStore.ts
```diff
@@ -9,3 +8,199 @@ export interface PersistentStore<T> {
...
+    this.crypto.subtle.generateKey(
+      {
+        name: "AES-GCM",
+        length: 256
+      },
+      true,
+      ["encrypt", "decrypt"]
+    ).then(result => {
+      this.masterKey = result;
+      this.wrapCryptoKey(this.masterKey);
```
wei-lee: this function does more than just generate the master key, it actually generates a key, and wrap it. So I suggest decouple the wrap part from this function. Chain them together in the init call to make the code easier to read.

packages/sync/src/PersistentStore.ts
```diff
@@ -9,3 +8,199 @@ export interface PersistentStore<T> {
...
+      const salt = this.strToArray(saltStr);
+      return salt;
+    }
+  }
+
+  // wrap the master key and store it under the "masterKey" key.
+  // when wrapped it can be stored safely and later unwrapped
+  private async wrapCryptoKey(keyToWrap: CryptoKey) {
+    // get the key encryption key
+    const keyMaterial = await this.getKeyMaterial(this.password);
```
wei-lee: I would prefer to have the password value to be passed into this function, will make it easier to test.

packages/sync/src/PersistentStore.ts
```diff
@@ -9,3 +8,199 @@ export interface PersistentStore<T> {
...
+    ).then(result => {
+      const resultAsString = this.arrayToStr(result);
+      this.storage.setItem("masterKey", this.arrayToStr(salt) + ":" + resultAsString);
+    });
+
+  }
+
+  // unwrap the safely stored masterKey from storage to be used for decryption
+  private async unwrapCryptoKey(masterKey: ArrayBuffer, salt: Uint8Array) {
+    const key = await this.getKeyMaterial(this.password);
```
wei-lee: same here, make the password a parameter of the function.

wei-lee:
@StephenCoady I have left some comments. Let's move it out and add some unit tests for it.

StephenCoady:
moved this to a package here: https://www.npmjs.com/package/web-encrypted-storage

if we want to I can move the repo under aerogear. I will create a PR to update the docs then if we do.

PR #327, opened 2019-04-12
Non blocking queue

wtrocki:
## Motivation

Avoid blocking offline queue. 
General idea is to return an error to the user instantly
and then trigger another mutation that will be enqueued as an offline one.

1. The request comes to an offline queue.
2. Return observable with error instantly (mutate().catch() will be triggered)
3. Attach observable to error object.

```javascript
client.mutate.try()
.catch(error => {
    error.offlineObservable.then((data)=>{
        // Successfully mutated after error
    })
})
```

Screen Shot 2019-04-12 at 5 56 41 PM

## Tasks

- [x] Implement nonblocking behavior
- [ ] Unify storage


packages/sync/src/links/LinksBuilder.ts
```diff
@@ -8,10 +8,10 @@ import { AuditLoggingLink } from "./AuditLoggingLink";
 import { DefaultMetricsBuilder, MetricsBuilder } from "@aerogear/core";
 import { LocalDirectiveFilterLink } from "./LocalDirectiveFilterLink";
 import { createUploadLink } from "apollo-upload-client";
-import { isMutation, isOnlineOnly, isSubscription, isMarkedOffline } from "../utils/helpers";
+import { isMutation, isOnlineOnly, isSubscription } from "../utils/helpers";
 import { defaultWebSocketLink } from "./WebsocketLink";
 import { OfflineLink } from "./OfflineLink";
-import { NetworkStatus } from "../offline";
+import { NetworkStatus, OfflineRestoreHandler } from "../offline";
 import { extensionsLink } from "./ExtensionsLink";
```
jhellar: This import can be removed.

packages/sync/src/links/OfflineLink.ts
```diff
@@ -37,11 +43,30 @@ export class OfflineLink extends ApolloLink {
   }
 
   public request(operation: Operation, forward: NextLink) {
-    if (!this.online || isMarkedOffline(operation)) {
+    // Reattempting operation that was marked as offline
+    if (OfflineRestoreHandler.isMarkedOffline(operation)) {
+      logger.log("Enqueueing offline mutation", operation.variables);
```
jhellar: `logger.log(...)` -> `logger(...)`

packages/sync/src/links/OfflineLink.ts
```diff
@@ -37,11 +43,30 @@ export class OfflineLink extends ApolloLink {
...
       return this.queue.enqueue(operation, forward);
     }
-    // We are online and can skip this link;
-    return forward(operation);
+
+    if (this.online) {
+      logger.log("Online: Forwarding mutation", operation.variables);
```
jhellar: `logger.log(...)` -> `logger(...)`

packages/sync/src/links/OfflineLink.ts
```diff
@@ -37,11 +43,30 @@ export class OfflineLink extends ApolloLink {
...
+      // We are online and can skip this link;
+      return forward(operation);
+    }
+
+    if (!this.offlineMutationHandler) {
+      logger.log("Error: Offline link setup method was not called");
```
jhellar: `logger.log(...)` -> `logger(...)`

packages/sync/src/links/OfflineLink.ts
```diff
@@ -37,11 +43,30 @@ export class OfflineLink extends ApolloLink {
...
+      return forward(operation);
+    }
+
+    const operationEntry = new OperationQueueEntry(operation, forward);
+    this.offlineMutationHandler.mutateOfflineElement(operationEntry);
+    return new Observable(observer => {
+      logger.log("Returning error to client", operation.variables);
```
jhellar: `logger.log(...)` -> `logger(...)`

packages/sync/src/offline/OfflineRestoreHandler.ts
```diff
@@ -47,27 +48,52 @@ export class OfflineRestoreHandler {
...
-        mutation: item.operation.query,
-        // Restore optimistic response from operation in order to see it
-        optimisticResponse,
-        // Pass client update functions
-        update: updateFunction,
-        // Pass extensions as part of the context
-        context: { extensions }
-      };
-      this.apolloClient.mutate(mutationOptions);
+      await this.mutateOfflineElement(item);
```
jhellar: `await` should probably not be there. When removed, integration tests pass.

wtrocki: Good catch!

StephenCoady:
@wtrocki took a look at the code and it looks great to me. I played with it too and its working great. When you make this ready for review I will review it fully!

packages/sync/src/offline/OfflineRestoreHandler.ts
wtrocki: Actually going to address that so we will have async here.

wtrocki:
@StephenCoady I'm going to record video on this and produce some diagrams to explain this idea. This will be the new architecture so the change will require major bump.

wtrocki:
Ready for testing but integration tests worry me now a lot.

wtrocki:
Ready for verification.
I have played with that on IOS and it works like a dream.
Outstanding part will be to write documentation. I will add documentation to separate repository (and release dev version now)

PR #331, opened 2019-04-17
fix: restore blocking/non blockin

wtrocki:
Diffrent idea that I'm exploring at the moment.


wtrocki:
This kinda introduces additional flags. 
I think that we should not be doing restore ourselves and block client creation.
So in long run users will control when they want to restore offline mutations (probably after cordova is ready etc.) This idea works but is kinda bad execution.

wtrocki:
Actually this is breaking observable chain. Second call always needs to be blocking

PR #339, opened 2019-04-29
Dynamic listeners 

wtrocki:
## Motivation

Enable `client.registerOfflineEventListener` to dynamically register global listeners. 

ping @StephenCoady


wtrocki:
I have messed up something with the git. Going to fix that.

wtrocki:
Ready for review. Going to improve api docs.

wtrocki:
@StephenCoady Applied fix for the problem that we have mentioned on call.

packages/sync/src/offline/OfflineMutationsHandler.ts
```diff
@@ -18,14 +18,12 @@ export const logger = debug.default(MUTATION_QUEUE_LOGGER);
...
-  private store: OfflineStore;
   private mutationCacheUpdates?: CacheUpdates;
 
-  constructor(apolloClient: ApolloClient<NormalizedCacheObject>, clientConfig: DataSyncConfig) {
-    this.apolloClient = apolloClient;
+  constructor(private store: OfflineStore,
+              private apolloClient: ApolloClient<NormalizedCacheObject>,
+              clientConfig: DataSyncConfig) {
     this.mutationCacheUpdates = clientConfig.mutationCacheUpdates;
-    this.store = new OfflineStore(clientConfig);
```
StephenCoady: shouldn't this assign from store in the constructor now?

wtrocki: contructor arguments with `private` clasiffier will get instantly an private field of the class. Is that what you mean?

StephenCoady: it certainly is 🤦‍♂

PR #341, opened 2019-04-30
WIP cache updates

wtrocki:
This is a first complete draft of the helpers for mutations and subscriptions. To test it, you can use this branch of the showcase application: https://github.com/aerogear/ionic-showcase/pull/171

### Usage:

**Mutations:**

```
const helperOptions: MutationHelperOptions = {
  // see interface
};
const mutationOptions = createMutationOptions(helperOptions);
apollo.mutate<T>(mutationOptions);
```

The create mutation options will then create some useful stuff like the optimistic response, the update function for the cache etc.

**Subscriptions:**

There are two ways to use this. the `createSubscriptionOptions` file exports a function of the same name and it also exports `subscribeToMoreHelper`.

To use createSubscriptionOptions you can do the following:

```
const myObservableQuery = apollo.watchQuery<T>({options});
const options = createSubscriptionOptions({
// see interface
});
myObservableQuery.subscribeToMore(options);
```

This can be either run multiple times for different SubscriptionOptions such as create, update, delete etc. or you can use the subscribeToMoreHelper function. This function takes two parameters. One is the myObservableQuery from above and the other is an array of `SubscriptionHelperOptions`. This function calls `subscribeToMore` for each set of options it receives.


wtrocki:
@StephenCoady Just had this idea. We may need to create subscription helper at some point to provide full story and deduplication feature out of the box. WDYT?

StephenCoady:
@wtrocki Really like it, I have followed this approach when starting the subscription helpers and based off this branch. I think I will have a first draft soon.

wtrocki:
@StephenCoady perfect!

StephenCoady:
@wtrocki this is usable now for our showcase app with the subscriptions. the code here: https://github.com/aerogear/ionic-showcase/blob/master/src/app/services/sync/item.service.ts#L46-L72

can now be replaced with this: 

```
    getTasks.subscribeToMore(createSubscriptionOptions(TASK_ADDED_SUBSCRIPTION, GET_TASKS, CacheOperation.ADD));
    getTasks.subscribeToMore(createSubscriptionOptions(TASK_UPDATED_SUBSCRIPTION, GET_TASKS, CacheOperation.UPDATE));
    getTasks.subscribeToMore(createSubscriptionOptions(TASK_DELETED_SUBSCRIPTION, GET_TASKS, CacheOperation.DELETE));
```

and everything should work the same. The only caveat is that I had to break our old subscription model of add, update and delete all living under the same subscription. It makes this difficult as each mutation will actually fire the helper regardless of the type of subscription. Something we should talk about probablyl

packages/sync/src/cache/CacheOperation.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * Cache update that involves removing item from the query
+   */
+  DELETE = "delete",
+
+  /**
+   * Cache update that involves updating an item from the query
+   */
+  UPDATE = "update"
```
wtrocki: @StephenCoady See REFRESH above. It served the same purpose.

StephenCoady: Excellent, didn't realise updates were handled automatically. Will remove update completely.

wtrocki: They are for mutation. Not sure if that works for subscriptions. 
Let's test it first. Just mentioning that we can reuse refresh or drop update.

StephenCoady: Yeah I tested it before I replied, it works 👍

packages/sync/src/cache/CacheOperation.ts
```diff
@@ -10,27 +10,21 @@
...
   /**
    * Cache update that involves adding new item to the query
    */
   ADD = "add",
-
   /**
-   * Cache update that involves removing item from the query
+   * Cache update that involves notifying cache that  specific query is refreshed.
+   * It can be applied for changes that perform updates on different objects etc.
+   * Used by default
```
wtrocki: I think ADD is default

wtrocki:
Closing to create new PR

PR #344, opened 2019-05-08
Storage poc

wtrocki:
Creating the same PR on separate branch as I cannot push to @StephenCoady fork.


wtrocki:
Executed performance tests:

- Latest release (no improvements in storage - using window.localstorage)
4mb of cache 7mb of offline data. The device works as a radiator

- IndexedDB without tunning
12 mb of cache made my device explode
8 mb of offline

- IndexedDB Object based (no JSON.stringify)
21mb of cache 
90mb of offline data 

No slow down in application (until the app is restarted ;)
The ultimate fix will be to add indexing/cursor instead of fetching entire offline data at startup however this will be noticeable with really large datasets.

Still working on this. 
Ping @StephenCoady

wtrocki:
Screenshot 2019-05-09 at 22 38 03
Cache is saved as object directly.

PR #345, opened 2019-05-09
[WIP] feat: auth integration tests

jhellar:
### Description

https://issues.jboss.org/browse/AEROGEAR-9146

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,6 +61,76 @@ jobs:
       - run: CI=true npm run release:prep
       - run: TAG=$CIRCLE_TAG npm run release:validate
       - run: TAG=$CIRCLE_TAG npm run publish
+  build_android:
+    working_directory: ~/aerogear
+    docker:
+      - image: circleci/android:api-28-node
+      - image: verdaccio/verdaccio
+        name: verdaccio_instance
+    steps:
+      - checkout
+        # publish npm packages to local verdaccio instance for testing
+      - run: npm set registry http://verdaccio_instance:4873/
+      - run: npx npm-cli-login -u test -p test -e test@example.com -r http://verdaccio_instance:4873
+      - run: git config --global user.email "test@example.com"
+      - run: git config --global user.name "test"
+      - run: npx lerna version patch --no-push --yes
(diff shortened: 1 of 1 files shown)


wtrocki:
@jhellar CircleCi allows this ORB's to modularize builds. Maybe we can have actual logic for integration in this repository and execute this here. 
Apart from that this looks sooo nice.

wtrocki:
@jhellar How this was build for PR? It should be building things on tag only right ?

jhellar:
@wtrocki looking at the config.yml, publishing code and testing in browserstack could be modularized with ORBs. Will send a commit for it.

For the tag only, that is only for real npm publishing, right?

wtrocki:
I meant that it looks like this code executes for PRs. Is that right ? Fine to approve this once we clear it

wei-lee:
@jhellar @wtrocki stupid question, what is `ORB`?

wei-lee:
@jhellar how long the tests will take?

wtrocki:
The integration adds ~8 minutes to build time. (see build options below).
It is acceptable for PR's but it looks like it is configured for tags only now

jhellar:
@wei-lee docs for the ORBs https://circleci.com/orbs/ I understand it that it is about config reusing

jhellar:
@wtrocki I think that the circleci config `filters, tags, only, /.*/` is confusing - if I understand it right, it just says that any tag should trigger the build. The way to disable builds for PRs is with `branches, ignore` - https://github.com/aerogear/aerogear-js-sdk/pull/345/files#diff-1d37e48f9ceff6d8030570cd36286a61R161

wtrocki:
My private preference will be to not run this for PR's as we do a lot of nonauth related tests and 8 minutes overhead is too much. +100 for prerelease builds!

wei-lee:
@wtrocki @jhellar yeah, that would be my preference as well. Given that the integration tests will test a few services rather than just one, maybe we can run it as part of a release pipeline instead. We can chat about it a bit more.

jhellar:
https://issues.jboss.org/browse/AEROGEAR-9270 -> closing this PR.

PR #348, opened 2019-05-13
Cache update helpers

StephenCoady:
This is a first complete draft of the helpers for mutations and subscriptions. To test it, you can use this branch of the showcase application: https://github.com/aerogear/ionic-showcase/pull/171

### Usage:

**Mutations:**

```
const helperOptions: MutationHelperOptions = {
  // see interface
};
const mutationOptions = createMutationOptions(helperOptions);
apollo.mutate<T>(mutationOptions);
```

The create mutation options will then create some useful stuff like the optimistic response, the update function for the cache etc.

**Subscriptions:**

There are two ways to use this. the `createSubscriptionOptions` file exports a function of the same name and it also exports `subscribeToMoreHelper`.

To use createSubscriptionOptions you can do the following:

```
const myObservableQuery = apollo.watchQuery<T>({options});
const options = createSubscriptionOptions({
// see interface
});
myObservableQuery.subscribeToMore(options);
```

This can be either run multiple times for different SubscriptionOptions such as create, update, delete etc. or you can use the subscribeToMoreHelper function. This function takes two parameters. One is the myObservableQuery from above and the other is an array of `SubscriptionHelperOptions`. This function calls `subscribeToMore` for each set of options it receives.


packages/sync/src/cache/CacheOperation.ts
```diff
@@ -0,0 +1,30 @@
...
+  /**
+   * Cache update that involves adding new item to the query
+   */
+  ADD = "add",
+  /**
+   * Cache update that involves notifying cache that  specific query is refreshed.
+   * It can be applied for changes that perform updates on different objects etc.
+   * Used by default
+   */
+  REFRESH = "refresh",
```
wtrocki: Do we need refresh for subscriptions?

wtrocki:
Since we moved towards the variable and doc for operation input we can also refactor 	`optimisticResponseHelper` to accept these types instead of strings. 

Then we can simply create another wrapper method like `offlineMutation()` where we will supply `update` and `optimisticResponse` parameters that are generated. 

UX FTW!

packages/sync/src/cache/CacheUpdates.ts
```diff
@@ -12,8 +13,19 @@ import { MutationUpdaterFn } from "apollo-client";
  *  createTask: () => {...}
  * }
  */
-interface CacheUpdates {
+export interface CacheUpdates {
   [key: string]: MutationUpdaterFn;
 }
 
-export default CacheUpdates;
+export type UpdateFunction = (array: [CacheItem], newItem?: CacheItem) => CacheItem[];
```
wtrocki: This seems to be defined in Apollo. We introduce here a lot of types that may be available in Apollo. I'm more curious why this is happening? Are they private/not exposed?

packages/sync/src/cache/createMutationOptions.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  let updateFunction: MutationUpdaterFn;
+
+  switch (opType) {
+    case CacheOperation.ADD:
+      updateFunction = (cache, { data }) => {
+        try {
+          if (data) {
+            let queryResult = cache.readQuery({ query, variables }) as any;
+            const operationData = data[operation];
```
wtrocki: This can be empty as well. We need to check if is not undefined.

packages/sync/src/cache/createMutationOptions.ts
```diff
@@ -0,0 +1,125 @@
...
+              queryResult = [result];
+            }
+            cache.writeQuery({
+              query,
+              variables,
+              data: queryResult
+            });
+          }
+        } catch (e) {
+          throw (e);
```
wtrocki: Caching and throwing the same error is pointless. Originally I left that empty
because it was crashing for me on Android on application startup for the first time when cache was empty)

packages/sync/src/cache/createMutationOptions.ts
```diff
@@ -0,0 +1,125 @@
...
+    operationType = CacheOperation.ADD,
+    idField = "id"
+  } = options;
+  const operationName = getOperationFieldName(mutation);
+  const optimisticResponse = createOptimisticResponse(operationName,
+    typeName,
+    variables,
+    operationType === CacheOperation.ADD);
+
+  const update = getUpdateFunction(operationName, idField, updateQuery, operationType);
```
wtrocki: How this works without passing variables?

packages/sync/src/cache/createMutationOptions.ts
```diff
@@ -0,0 +1,125 @@
...
+    default:
+      updateFunction = () => {
+        return;
+      };
+  }
+  return updateFunction;
+};
+
+// this function takes a Query and returns its constituent parts, if available.
+const deconstructQuery = (updateQuery: Query): QueryWithVariables => {
```
wtrocki: [Minor] It will be cool to have those helpers in separate file.

packages/sync/src/cache/createSubscriptionOptions.ts
```diff
@@ -0,0 +1,94 @@
...
+        if (!newItem) {
+          return [...prev];
+        } else {
+          return [...prev.filter(item => {
+            return item[idField] !== newItem[idField];
+          }), newItem];
+        }
+      };
+      break;
+    case CacheOperation.REFRESH:
```
wtrocki: [Question] So for subscriptions we need refresh method?
What I'm worried that cache operations are common for mutation and subscriptions so people may pass refresh to mutate etc. Let's chat about options here before we apply anything because it may be not worth it.

packages/sync/test/CacheHelpersTest.ts
```diff
@@ -0,0 +1,35 @@
+import { expect, should } from "chai";
```
wtrocki: ❤️

packages/sync/src/cache/CacheOperation.ts
StephenCoady: Yeah I was wrong initially saying we didn't, we do

packages/sync/src/cache/CacheUpdates.ts
StephenCoady: Hmm. I saw the mutation updater function, but this is different. i didn't see anything that matched this type - i'll take another look

packages/sync/src/cache/createMutationOptions.ts
StephenCoady: updateQuery is a type with the Query and variables built in. i deconstruct that type to get the query and variables then

packages/sync/src/cache/createSubscriptionOptions.ts
StephenCoady: Yeah we need refresh for subscriptions. I've thought of that. I pass update but if you look here: https://github.com/aerogear/aerogear-js-sdk/blob/cache-update-helpers/packages/sync/src/cache/createMutationOptions.ts#L98 I have a default which returns no update function for an update.

I use it in the showcase here to make the point that it doesn't break anything: https://github.com/aerogear/ionic-showcase/blob/breakup-subscriptions/src/app/services/sync/item.service.ts#L78

packages/sync/src/cache/CacheUpdates.ts
wtrocki: Ignore that!

packages/sync/src/cache/createMutationOptions.ts
wtrocki: This is very confusing. Why not support the same API as client.mutation where they can supply query and variables as separate fields?

packages/sync/src/cache/createSubscriptionOptions.ts
wtrocki: @StephenCoady - Let's update comment that refresh is not default and it is used for subscriptions only then. OR rename `Subscription_UPDATE` ?

packages/sync/src/cache/createMutationOptions.ts
StephenCoady: this is the query you want to call to refresh the cache. 

so for example at the top level you call createMutationOptions with your mutation, variables, and updateQuery. The variables I listed there are the mutation variables. The one contained within the updateQuery are any variables you might want to pass with your cache query.

wtrocki: Right so I was confused all the way. Variables for mutation have no impact here but we need to have separate container for query variables. Sorry for that.

wtrocki:
Verified on showcase using following cases:
Offline flow (when the optimistic response is added)
- Creation
- Updates
- Deletion
- App restart

Data replication
- Creation
- Updates
- Deletion
- App restart

It worked for all the cases apart from deletion. For deletion, the device did not actually added anything to offline queue. It may be related with some unrelated bug but is best to check that out.

StephenCoady:
@wtrocki this should be fixed now. It was the optimistic response checking the wrong operation types.

PR #349, opened 2019-05-13
fix: Allow optional alias

wtrocki:


packages/push/src/PushRegistration.ts
```diff
@@ -78,6 +78,9 @@ export class PushRegistration {
       "alias": alias,
```
danielpassos: This line should not be removed in this case?

wtrocki:
@danielpassos I was just showing the problem. Feel free to take this over.

PR #359, opened 2019-05-16
fix: add clientId to auth payload delivered by subscriptions

darahayes:
##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/sync/src/links/WebsocketLink.ts
+++ b/packages/sync/src/links/WebsocketLink.ts
@@ -9,8 +9,8 @@ export const defaultWebSocketLink = (userOptions: DataSyncConfig, config: WebSoc
       // Params that can be used to send authentication token etc.
       connectionParams: async () => {
         if (userOptions.authContextProvider) {
-          const { header } = await userOptions.authContextProvider();
-          return { Authorization: header };
+          const { header, clientId } = await userOptions.authContextProvider();
+          return { ...header, clientId };
         }
       },
       connectionCallback: options.connectionCallback,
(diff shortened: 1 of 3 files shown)


packages/sync/src/links/WebsocketLink.ts
```diff
@@ -9,8 +9,8 @@ export const defaultWebSocketLink = (userOptions: DataSyncConfig, config: WebSoc
       // Params that can be used to send authentication token etc.
       connectionParams: async () => {
         if (userOptions.authContextProvider) {
-          const { header } = await userOptions.authContextProvider();
-          return { Authorization: header };
+          const { header, clientId } = await userOptions.authContextProvider();
+          return { ...header, clientId };
```
wtrocki: [Idea] We do decompose that header on the server to token so maybe we can pass token directly instead of doing that?

darahayes: By having the server expect this format, we do not break compatibility with GraphQL playground/GraphiQL/Altair. Remember that `/token` endpoint we use to quickly copy paste that JSON object containing the header into the playground? The playground will also pass that same headers object in the connectionParams. If we accepted a different format then the playground would authenticate for regular queries but not for subscriptions. Plenty of other people doing it this way too.

wtrocki: Right. I forgot that you have explained that before. My bad!

wtrocki:
Going to publish that as dev release now

PR #362, opened 2019-05-20
Subscribe alias topic

danielpassos:
When app is running on Android devices, subscribe the it in the Firebase Cloud Messaging topics using the UnifiedPush Server VariandID, and the categories used in the registration process. That is how UnifiedPush Server send message to the Android devices using no criteria and a specifically category name.

PS: There is more work to be done like add a unregistered method but it will be done in another issues/PR

Related Issue:

* [AEROGEAR-9176](https://issues.jboss.org/browse/AEROGEAR-9176)

## Screenshots

**Sending message for all devices**

Screenshot 2019-05-20 09 50 20

**Sending message for an specifically alias**

Screenshot 2019-05-20 09 50 45

**Sending message for an specifically category** 

Screenshot 2019-05-20 09 51 08

This requires [this PR](https://github.com/aerogear/ionic-showcase/pull/182)

### DIFF
--- a/packages/push/src/PushRegistration.ts
+++ b/packages/push/src/PushRegistration.ts
@@ -37,7 +37,7 @@ export class PushRegistration {
    * @param categories array list of categories that device should be register to.
    */
   public register(deviceToken: string, alias: string = "", categories: string[] = []): Promise<void> {
-    if (!window || !window.device) {
+    if (!window || !window.device || !window.PushNotification) {
       return Promise.reject(new Error("Registration requires cordova plugin. Verify the " +
         "@aerogear/cordova-plugin-aerogear-push plugin is installed."));
     }
--- a/packages/push/src/PushRegistration.ts
+++ b/packages/push/src/PushRegistration.ts
@@ -1,4 +1,3 @@
-
 import { isCordovaAndroid, isCordovaIOS, ServiceConfiguration, ConfigurationService } from "@aerogear/core";
 import axios from "axios";
 
@@ -18,7 +17,8 @@ export class PushRegistration {
   public static readonly TYPE: string = "push";
   public static readonly API_PATH: string = "rest/registry/device";
 
-  private pushConfig?: ServiceConfiguration;
+  private readonly pushConfig?: ServiceConfiguration;
+  private _objectInstance: any;
 
   constructor(config: ConfigurationService) {
     const configuration = config.getConfigByType(PushRegistration.TYPE);
@@ -41,12 +41,15 @@ export class PushRegistration {
       return Promise.reject(new Error("Registration requires cordova plugin. Verify the " +
         "@aerogear/cordova-plugin-aerogear-push plugin is installed."));
     }
+
     if (!this.pushConfig || !this.pushConfig.config || !this.pushConfig.url) {
       return Promise.reject(new Error("UPS registration: configuration is invalid"));
     }
+
     if (!deviceToken) {
       return Promise.reject(new Error("Device token should not be empty"));
     }
+
     let platformConfig;
     const url = this.pushConfig.url;
     if (isCordovaAndroid()) {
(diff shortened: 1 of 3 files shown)


packages/push/src/PushRegistration.ts
```diff
@@ -61,14 +64,28 @@ export class PushRegistration {
...
     if (!variantId) {
       return Promise.reject(new Error("UPS registration: variantId is not defined."));
     }
+
+    const variantSecret = platformConfig.variantSecret;
     if (!variantSecret) {
       return Promise.reject(new Error("UPS registration: variantSecret is not defined."));
     }
+
+    this._objectInstance = window.PushNotification.init(
```
wtrocki: This looks like registration plugin getting direct connection to cordova plugin. 
Idea was that no matter what platform it will be it will be just doing the registration.
Just mentioning.

danielpassos: Hrm, so... how to handle FCM topic subscriptions?

wtrocki: I'm just saying that this package was purely to do UPS registrations. 
FCM and other stuff is Phonegap Push Plugin related so bringing this here ties this entire package to PhoneGap push. 

It is nice to get helpers for this stuff, so maybe we can add docs and showcase integration as well. 

The entire helper is really basic and it is a calling subscribe method so maybe showcase should have this and we should document that.

Feel free to ignore me here.

wtrocki: Yes. Ignore me here. I have already approved changes.
Let's fix the build to do verification

danielpassos: I got your point and totally agree with you. The only point is without this the app will not be able to receiver message from UPS. Only if the message is sending directly to the device using an alias.

TL:DR when you send a message on UPS without any criteria it will use FCM topic to delivery a message for all devices registered on the topic (with the same name/id of the variant) same for the category.

If you understood you correctly, what you are recommending is to move that code/feature to the `push-cordova` like a helper method and document it's needed to be fired if the app is running on Android devices. Am I correct?

wtrocki: I think is ok to have it this way. Alternatively we can have a strategy that will be default assume and use window.PushNotification.register is default one (if exist) and user can pass something to constructor. That way we provide the same functionality but people could still use let's say old plugin or anything they will need to use (like web) etc.

danielpassos:
Test fixed and rebased!

danielpassos:
@wtrocki Actually we already have this kind of test => https://github.com/aerogear/aerogear-js-sdk/blob/master/packages/push/src/PushRegistration.ts#L40-L43

```js
if (!window || !window.device) {
      return Promise.reject(new Error("Registration requires cordova plugin. Verify the " +
        "@aerogear/cordova-plugin-aerogear-push plugin is installed."));
}
```

Do you wanna I improve this testing plugin checking or replace it to test if the `window.PushNotification` exists?

danielpassos:
@wtrocki I have added one more check

packages/push/src/PushRegistration.ts
darahayes: @wtrocki is right about the original intention of this library which was to provide UPS registration only and not be coupled to the PhoneGap plugins. This would allow this library to be portable to the browser and to other platforms like React Native. However I also get @wei-lee and @danielpassos's point about making it easier to perform the actual registration with firebase.

One of the problems with the current approach in this PR is that the `register` function will now fail (silently) in all platforms except on a device. I think this approach is not optimal.

Instead I think we should provide the PhoneGap push registration in a separate function that is not called immediately when a user calls `register()`. 

I think the usage in a user's app should be something like this.

```js
const push = new PushRegistration(config) // the name of this class kinda sucks as well 😄
await push.registerUPS(token, alias, categories) // register with UPS only

try {
  await push.registerFirebaseTopics(categories) // perform the extra phonegap specific registration
} catch(e) {
  // handle error. Could be because wrong platform or could be a legit error.
}

File: packages/push/src/PushRegistration.ts [hunk 2]
```diff
@@ -78,13 +95,29 @@ export class PushRegistration {
...
-      return instance.post(PushRegistration.API_PATH, postData).then(() => resolve()).catch(reject);
+      return instance.post(PushRegistration.API_PATH, postData)
+      .then(
+        () => {
+          if (isCordovaAndroid()) {
+            this.subscribeToFirebaseTopic(variantId);
+            for (const category of categories) {
+              this.subscribeToFirebaseTopic(category);
+            }
+            resolve();
```
darahayes: A slightly nicer and more readable way to structure that would be as follows

```
if (isCordovaAndroid()) {
  // do the things
}
resolve()
```

there isn't really any need for the else statement and generally, it's not a great pattern to have multiple `resolve()` calls in the code because it can be confusing.

danielpassos: Done!

PR #364, opened 2019-05-21
fix: expose update function

StephenCoady:
### Description

This PR does two things.

1. Need to expose the getUpdateFunction so that a developer can manually generate cache update functions. Main usage is for storing cache update functions across device restarts.
see: https://github.com/aerogear/ionic-showcase/pull/183/files#diff-f8726a2856d5cf9931bf35cf2e8280a4R15 

2. It changes the interface so that `createMutationOptions` can now accept an array of type `Query` as well as a single Query.

The function signature is now one of the following:

```
createMutationOptions({
  mutation: DocumentNode,
  variables: OperationVariables,
  updateQuery: { query: GET_TASKS, variables: {...} },
  typeName: string,
  operationType: CacheOperation,
  idField: string
});

createMutationOptions({
  mutation: DocumentNode,
  variables: OperationVariables,
  updateQuery: [ 
     { query: GET_TASKS, variables: {...} } ,
     { query: OTHER_QUERY, variables: {...} } 
  ],
  typeName: string,
  operationType: CacheOperation,
  idField: string
});
```
Docs PR in progress: https://github.com/aerogear/mobile-docs/pull/699

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/sync/src/cache/createMutationOptions.ts
+++ b/packages/sync/src/cache/createMutationOptions.ts
@@ -37,10 +37,12 @@ export const createMutationOptions = (options: MutationHelperOptions): MutationO
   const update: MutationUpdaterFn = (cache, { data }) => {
     if (isArray(updateQuery)) {
       for (const query of updateQuery) {
-        getUpdateFunction(operationName, idField, query, operationType)(cache, { data });
+        const updateFunction = getUpdateFunction(operationName, idField, query, operationType);
+        updateFunction(cache, { data });
       }
     } else {
-      getUpdateFunction(operationName, idField, updateQuery, operationType)(cache, { data });
+      const updateFunction = getUpdateFunction(operationName, idField, updateQuery, operationType);
+      updateFunction(cache, { data });
     }
   };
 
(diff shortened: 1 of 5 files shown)


packages/sync/src/cache/createMutationOptions.ts
```diff
@@ -40,7 +40,7 @@ export const createMutationOptions = (options: MutationHelperOptions): MutationO
 // returns the update function used to update the cache by the client
 // ignores the scenario where the cache operation is an update as this is handled automatically
 // from Apollo client 2.5 onwards.
-const getUpdateFunction = (
+export const getUpdateFunction = (
```
wtrocki: API Documentation 💭 ?

wtrocki:
PR seems to be doing much more. Update of the description may be required.

packages/sync/src/cache/createMutationOptions.ts
```diff
@@ -33,14 +34,29 @@ export const createMutationOptions = (options: MutationHelperOptions): MutationO
     typeName
   });
 
-  const update = getUpdateFunction(operationName, idField, updateQuery, operationType);
+  const update: MutationUpdaterFn = (cache, { data }) => {
+    if (isArray(updateQuery)) {
+      for (const query of updateQuery) {
+        getUpdateFunction(operationName, idField, query, operationType)(cache, { data });
```
wtrocki: I would split gettingFunction and execution to two lines. 
More readable.

PR #373, opened 2019-05-30
Conflict resolution with full access to diff (base)

wtrocki:
##  Background

Having access to base allows us to get access to the diff if the conflict happens. 
Having a fast and reliable mechanism to access to base objects is essential.

## Approaches

Conflicts could be resolved on the client and server depending on the developer choice however depending on where conflict is resolved we might have different capabilities to perform diffing and overall conflict resolution.

### Provide diff mechanism on the server by persisting previous state

Persist the previous state of the data so server-side version can be compared against it giving diff.
When conflict will happen server or client will have access to base object that will allow users to perform 3 way merge using any strategy. This way we can also provide diffs for the client that can also provide strategies or even resolve the client visually (same as when doing git merge)

Alternatively to storing entire objects, we can persist diffs, however, this will make the comparison process more complex.

When resolving conflicts changes are compared against base giving nice diff for server and client.
This approach is already implemented, however, we require users to handle persistence for previous states, 
which is an additional overhead.

#### Pros

- Enables tracking changes and auditing them
- Enables full flexibility in terms of conflict resolution

#### Cons

- Requires separate storage (however we can provide helpers to handle those later)

### Record limited historical data on client before change happens

Instead of saving every edit on server we can record changes on each individual client. If a conflict happens then we can use that information to calculate diffs. 
This reduces the amount of data that needs to be saved and gives simplicity to end users.
 
##### Pros

- Works almost out of the box on server
- Handling logic is very simple to understand and implement for developers

## Solution in this PR

This PR implements approach nr2 - storing history on client.
In this PR we are trying to keep base information in-memory and going to persist it for offline changes only.

basechanges

PR is going to introduce BaseState link that will be responsible for adding and removing base data from memory. For offline changes, those will be also persisted in a local store (IndexedDB). Conflict link will be extended to take base information into consideration and calculate diff.

Screenshot 2019-05-30 at 21 18 36

# Tasks

- [x] Provide in the memory store for base
- [x] Implement base link
- [x] Implement persistence
- [x] Extend Conflict link
- [ ] Extend integration tests
- [ ] Documentation


wtrocki:
@StephenCoady @darahayes This has now all required abstractions to start work.
Feel free to chime in and claim tickets as I will be looking to get that sorted next week.

packages/sync/src/links/LinksBuilder.ts
```diff
@@ -74,7 +76,9 @@ export const defaultHttpLinks = async (config: DataSyncConfig, offlineLink: Apol
   }
 
   if (config.conflictStrategy) {
-    links.push(conflictLink(config));
+    const provider = new BaseStateStore();
+    links.push(new BaseStateLink(provider));
```
wtrocki: Probably needs to be a composite link that both interact with base state store.

packages/sync/test/OfflineClientTest.ts
```diff
@@ -31,4 +34,26 @@ describe("Top level api tests", () => {
     initClient.registerOfflineEventListener(
       new CompositeQueueListener({ queueListeners: [] }));
   });
+
+  it("check direct cache read", async () => {
+    const client = new OfflineClient({ httpUrl: url, storage, networkStatus });
+    const initClient = await client.init();
+    // const data = {
```
wtrocki: Remove this

wtrocki:
This PR may be also nice to follow:
@Lavanyagaur22 @ankitjena

packages/sync/src/conflicts/base/BaseStateLink.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  constructor(private baseState: BaseStateProvider) {
+    super();
+  }
+
+  public request(operation: Operation, forward?: NextLink) {
+    // TODO get data from cache
+    // TODO filter only for mutations (EDITS)
+    // TODO Add persistence depending if offline or not
+    this.baseState.save({}, operation.toKey(), false);
```
wei-lee: what is the return value for `operation.toKey()`, will it be unique for each version of the base state object? The reason I am asking is thinking about what will happen if when mutation A is in flight and waiting for response from the server, mutation B happened on the same object? Will it override the base state for mutation A if the key is the same?

packages/sync/src/conflicts/base/BaseStateLink.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (!forward) {
+      return null;
+    }
+    return forward(operation).map((data) => {
+      this.baseState.delete(operation.toKey(), false);
```
wei-lee: same here, looks like the return value of the key has to be unique for each base state object?

wei-lee:
@wtrocki nice proposal, looks very promising and simple to understand. The only potential problem I can see is that it could increase the size of the storage required by the app depending on how many mutations are there. But it should not be a big deal given that we will remove them once the mutations are applied. 

I also added some comments around the key value for the base state object, I think they might need to be unique for each version.

packages/sync/src/conflicts/base/BaseStateLink.ts
wtrocki: Very valid point! We hav
This is WIP so some of the code elements are still not solid enough.
We have unique keys we generate for offline etc.

The `operation.toKey()` is used temporarily. Key will contain hash, passed variables etc. so it will be unique itself. The problem we facing now is to actually access base data. To access base data we will need to know the actual type we request from server and this is known only when request is finished. This Type is not available in Apollo-client for javascript (it is well known in Android and IOS). To overcome that we can:

- force application wide id's in a cache (nah.. this is to restrictive and limiting for current integrations)
- require type information to be passed when mutating 
This can be done by:
a) Directive - @returnType(name='Task')
b) Simply reusing cache update helpers that currently also require type information
- Another idea.

We trying our best to avoid storing entire history on the server, but there will be number of challenges to overcome and there will be always tradeof. 

To provide more context. We will get type information from server after the successful request but then this is too late.

We could technically access base when conflict happened (after request) but after investigating this case we found out that base can often be overridden by another query, subscription or even previous mutation from offline store so this method will not be 100% reliable and in some cases we will need to fall back to conflict resolution without diff which is not desired. 

Entire PR is still in the progress but once we figure out how to access cache data before sending request to server we will be in a very good position to resolve this problem without any corner cases and fallbacks to non diff resolutions.

wtrocki:
> The only potential problem I can see is that it could increase the size of the storage required by the app depending on how many mutations are there.

This is very clever as everything will be in memory and it will be gone once request is finished so storage is not affected. 

We going to save those only when we know that we are offline and this just adds a little bit over the currently saved objects so not as big impact as holding entire copy of cache for history purposes (which was my first idea)

packages/sync/src/conflicts/base/BaseStateLink.ts
darahayes: > require type information to be passed when mutating
> This can be done by:
> a) Directive - @returnType(name='Task')
> b) Simply reusing cache update helpers that currently also require type information

I like that idea. Just looking at it from a slightly different perspective. Could it be possible to include a client side directive, something like `conflictsOn(TYPE: String)`?

The purpose of this directive would be for developers to explicitly say 'This mutation may cause a conflict on this particular type'. The name here is generic enough that if we need to add additional logic in there to handle conflicts in the future then that would be possible.

wtrocki:
Update from the call.
Directive-based approach may be preferable, but we can spike both approaches and demo them

wtrocki:
@StephenCoady are you fine for me to remove merge commits (they make things nasty later)

StephenCoady:
@wtrocki sure, i should have done it manually. always forget. sorry

wtrocki:
Top level impact:

- Server side conflict resolution is going to be removed as we cannot provide enough data on the server to resolve conflict (it will require aproach nr1 which is too expensive to implement on user side)

- Client side is going to record conflict data as base. 
Conflict link will the first in the chain and can check if data that is being sent to server is actual

Screenshot 2019-06-10 at 13 21 11

- Additional type required will be passed thru context object (and added to offlineMutation helper.

wtrocki:
@StephenCoady Very rough state. Testing that now

packages/sync/src/conflicts/base/BaseStateProvider.ts
```diff
@@ -0,0 +1,47 @@
+
+/**
+ * Represents any object in the cache before applying modification from
+ * currently processed mutation.
+ *ß
```
wtrocki: :)

wtrocki:
This will require us to provide type in the mutation for showcase app in order to test it.

packages/sync/src/conflicts/strategies/strategies.ts
```diff
@@ -0,0 +1,37 @@
...
+
+
+// Used as default strategy for SDK
+export const clientWins: ConflictResolutionStrategy =
+  (base: ConflictResolutionData, server: ConflictResolutionData, client: ConflictResolutionData) => {
+    const { clientDiff } = calculateDiffs(base, client, server);
+    return Object.assign(base, clientDiff);
+  };
+
+function calculateDiffs(base: any, client: any, server: any) {
```
wtrocki: Very rough state. We need to test it

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -1,15 +1,151 @@
...
+      this.config.baseState.save(base, operation.toKey(), false);
+    }
+  }
+
+  private conflictErrorHandler(errorResponse: ErrorResponse) {
+    const { response, operation, forward, graphQLErrors } = errorResponse;
+    const data = this.getConflictData(graphQLErrors);
+    if (data && this.strategy) {
+      let resolvedConflict;
+      const base = getObjectFromCache(this.cache, operation.variables.id, data.returnType);
```
wtrocki: TODO: We need to handle if returnType is not provided

wtrocki:
Our conflict implementation works only on root level. Any nested (array like) mutations are very very rare case and we should document that this is not supported. Conflicts will be reported only on root level due to fact that we do not have way to track all relationships and save their bases for request.

FYI @darahayes @StephenCoady

wtrocki:
I got this working quite nicely - however we are hitting problem where server side id is number but data in cache (id particularly is saved as string). 
This obviously bring problems when comparing those are not the same object despite having the same value

wtrocki:
We will need to focus now on server side changes to resolve issues around diferent data set. 

We should be sending fields from server side but only those that were provided by client etc.

wtrocki:
I think if we go with string representation of any data comming from server we cannot have it wrong. Transforming any field to string will work quite nicely. 

Additionally, we will need to filter out metadata fields (version). With current implementation, the version is always causing hard conflicts on server side. 

I think we no longer can maintain the current interface for resolution. 
We need to perform diff calculation beforehand and notify/call api only on hard conflicts. This minor changes will be quite cool to have.

StephenCoady:
@wtrocki I can take a look at the server side changes to bring what we need in to line with what is here. I think we can get around the version aspect by just adding the diff to the state interface? That way each state implementation knows what fields it should/shouldn't diff.

wtrocki:
@StephenCoady - I think we should do proper planning today with @darahayes.
Let's sort out some time

packages/sync/src/conflicts/base/BaseStateProvider.ts
```diff
@@ -0,0 +1,47 @@
...
+export interface BaseStateProvider {
+
+  /**
+   * Add new base to currently
+   *
+   * @param base - base state data
+   * @param key - key under data should be persisted
+   * @param persist flag used to check if data should be persisted to queue
+   */
+  save(base: BaseState, key: string, persist?: boolean): Promise<void>;
```
wtrocki: Persist should probably be separate method

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -1,15 +1,152 @@
...
+      return forward(operation);
+    }
+    this.processBaseState(operation, forward);
+    // FIXME removing base state
+    return this.link.request(operation, forward);
+  }
+
+  private processBaseState(operation: Operation, forward: NextLink) {
+    const context = operation.getContext();
+    const base = getObjectFromCache(this.cache, operation.variables.id, context.returnType);
```
wtrocki: TODO: This is currently happening twice for offline changes. We need to figure out how to get it only once and persist it on the other chance. operation.getContext().isOffline is good marker here (we can inject that to this link really)

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -1,15 +1,152 @@
...
+    if (base && Object.keys(base).length != 0) {
+      const currentChange = this.stater.currentState(base);
+      const userState = this.stater.currentState(operation.variables);
+      if (currentChange !== userState) {
+         // 🙊 Input data is conflicted with the latest server projection
+        throw new LocalConflictError(base, operation.variables);
```
wtrocki: This is actually not going to work since we should be returning observable with error. 
THrowing error here breaks js stack

packages/sync/src/conflicts/strategies/strategies.ts
StephenCoady: should the strategy just return a boolean of resolved? thats an easy way to not throw the error in the conflict link

packages/sync/src/conflicts/strategies/strategies.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+  const clientDiff: any = {};
+  for (const key of Object.keys(client)) {
+    if (base[key] && base[key] !== client[key]) {
+      clientDiff[key] = client[key];
+    }
+  }
+  const serverDiff: any = {};
+  for (const key of Object.keys(client)) {
+    if (base[key] && base[key] !== server[key]) {
```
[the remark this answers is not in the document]
wtrocki: How about pluggability? :>

wtrocki:
> should the strategy just return a boolean of resolved? thats an easy way to not throw the error in the conflict link 

This entire area needs to be rewritten. We need to separate between hard and soft conflicts. In both cases we sending data to server but hard one will be available for people to solve while soft one will be done automatically. 

Currently, we are using the old structure which do not really work.
This is no where to be considered final solution. Just an hack to try things in showcase end to end and see the problems.

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -1,15 +1,152 @@
...
+      return forward(operation);
+    }
+    const currentState = this.stater.currentState(operation.variables);
+    if (!currentState) {
+      // 🙈 Missing state information in request.
+      // No action here
+      return forward(operation);
+    }
+    this.processBaseState(operation, forward);
+    // FIXME removing base state
```
wtrocki: @StephenCoady THere are FIXME all over the place to implement things. 
Removing saving/base is not critical for the moment to prove this works and can be done over the time.

packages/sync/src/conflicts/strategies/strategies.ts
wtrocki: It works but it is not the final state.

packages/sync/src/conflicts/base/BaseStateStore.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Store that provides access to base state along with local persistence.
+ */
+export class BaseStateStore implements BaseStateProvider {
+  public baseState: any = {};
+  private storage: Store;
+
+  constructor() {
+      this.storage = new Store("base-store", "base-data");
```
darahayes: I think it would make sense to pass the storage implementation ("idb-localstorage") into the constructor here. It means we will need to define some basic interface for the storage but it will make this entire file really easy to unit test

packages/sync/src/conflicts/conflictLink.ts
```diff
@@ -1,15 +1,152 @@
...
+   */
+  baseState: BaseStateProvider
+}
+
+/**
+ * Conflict handling link implementation that provides ability to determine 
+ */
+export class ConflictLink extends ApolloLink {
+  private link: ApolloLink;
+  private stater: ObjectState;
```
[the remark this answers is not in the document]
darahayes: `stateManager` or `stateTracker` maybe?

packages/sync/src/conflicts/base/BaseStateStore.ts
wtrocki: Yeah. We need to abstract this. Another TODO :)

wtrocki:
Requires rebase

StephenCoady:
@wtrocki if its ok with you I will update this branch - there are a tonne of changes on my branch (including rebases) so it might make sense to push here. once we have everything in one place we can discuss whats left.

StephenCoady:
Just a note about the current state of this. It works at this point with the happy path. It needs refinement and cleaning up from here but the flow is now different.

No new storage - instead, the base is saved _with_ the operation and passed through links in its **context**. The persistence is now handled in a storage mechanism we already have available - the offline store.

The orchestration of the links is now as follows:

Base Link -- records the base state and saves it to the operation's context
|
Offline Link -- no changes here
|
Conflict Link -- now has the base available and can act accordingly. Employs a conflict handler to tie everything together.

TO DO:

- list of keys to ignore in the conflict handler is currently hardcoded. this should be pluggable
- currently we only have access to the conflict once, as a second error on the same operation will not be caught by the link twice. we should discuss options here.
- I have changed the structure of the default conflict strategy to no longer be a dictionary. it is now just a single strategy which can be used globally. to provide per mutation strategies developer's can  add a `strategy` function to each mutation's context.
- testing

The current happy path is:

1. Go offline client 1
1. Mutate title on client 1
1. Mutate description on client 2
1. Go online client 1
1. Happy scenario is now on server with title from client 1 and description from client 2

packages/sync/src/conflicts/VersionConflictHandler.ts
```diff
@@ -2,39 +2,41 @@ import { ConflictResolutionData } from "./strategies/ConflictResolutionData";
...
     this.checkConflict(this.options.base, this.options.client, this.options.server);
   }
 
   public checkConflict(base: ConflictResolutionData, client: ConflictResolutionData, server: ConflictResolutionData) {
     for (const key of Object.keys(client)) {
-      if (base[key] && base[key] !== client[key]
-      && !this.options.ignoredKeys.includes(key)) {
-        this.clientDiff[key] = client[key];
-      }
+      if (base[key] && base[key] !== client[key] && !this.ignoredKeys.includes(key)) {
```
wtrocki: very complex if :)

packages/sync/src/conflicts/VersionConflictHandler.ts
```diff
@@ -2,39 +2,41 @@ import { ConflictResolutionData } from "./strategies/ConflictResolutionData";
...
   public conflicted: boolean;
   public clientDiff: any = {};
   public serverDiff: any = {};
   public options: HandlerOptions;
+  private ignoredKeys: string[];
 
   constructor(options: HandlerOptions) {
     this.options = options;
     this.conflicted = false;
+    this.ignoredKeys = this.options.ignoredKeys || ["version", "id"];
```
wtrocki: This is the only place we put version. Why we really want to have this as versioned. The entire implementation is generic and this makes things much more complex for the end dev.

StephenCoady:
Closing this PR as it has gotten difficult to manage changes with the large amount of commits. going to squash some of the smaller commits and combine them in a new PR which will be easier to review.

PR #379, opened 2019-06-13
Add push unregister

danielpassos:
Add a method to unregistered a device on unified push server

You can use the ionic-showcase [AEROGEAR-9230](https://github.com/danielpassos/ionic-showcase/tree/AEROGEAR-9230) branch to test it

Register/Unregister screenshot

### DIFF
--- a/packages/push/src/PushRegistration.ts
+++ b/packages/push/src/PushRegistration.ts
@@ -25,7 +25,7 @@ export class PushRegistration {
 
   private readonly validationError?: string;
   private readonly variantId?: string;
-  private readonly http?: AxiosInstance;
+  private readonly httpClient?: AxiosInstance;
   private readonly push?: any;
 
   constructor(config: ConfigurationService) {
@@ -80,7 +80,7 @@ export class PushRegistration {
 
       if (!this.validationError) {
         const token = window.btoa(`${this.variantId}:${variantSecret}`);
-        this.http = axios.create({
+        this.httpClient = axios.create({
           baseURL: unifiedPushServerURL,
           timeout: 5000,
           headers: {"Authorization": `Basic ${token}`}
@@ -130,8 +130,8 @@ export class PushRegistration {
     };
 
     return new Promise((resolve, reject) => {
-      if (this.http) {
-        return this.http.post(PushRegistration.API_PATH, postData)
+      if (this.httpClient) {
+        return this.httpClient.post(PushRegistration.API_PATH, postData)
         .then(
           () => {
             const storage = window.localStorage;
@@ -172,9 +172,9 @@ export class PushRegistration {
     }
 
     return new Promise((resolve, reject) => {
-      if (this.http) {
+      if (this.httpClient) {
         const endpoint = PushRegistration.API_PATH + "/" + deviceToken;
-        return this.http.delete(endpoint, {})
+        return this.httpClient.delete(endpoint, {})
         .then(() => resolve())
         .catch(reject);
       } else {
--- a/packages/push/src/PushRegistration.ts
+++ b/packages/push/src/PushRegistration.ts
@@ -85,18 +85,18 @@ export class PushRegistration {
           timeout: 5000,
           headers: {"Authorization": `Basic ${token}`}
         });
-      }
 
-      this.push = window.PushNotification.init(
-        {
-          android: {},
-          ios: {
-            alert: true,
-            badge: true,
-            sound: true
+        this.push = window.PushNotification.init(
+          {
+            android: {},
+            ios: {
+              alert: true,
+              badge: true,
+              sound: true
+            }
           }
-        }
-      );
+        );
+      }
 
     } else {
       console.warn("Push configuration is missing. UPS server registration will not work.");
--- a/packages/push/src/PushRegistration.ts
+++ b/packages/push/src/PushRegistration.ts
@@ -120,6 +120,80 @@ export class PushRegistration {
   }
 
   /**
+   * Register deviceToken for Android or IOS platforms
+   *
+   * @param deviceToken token that will be sent to Unified Push server
+   */
+  public unregister(deviceToken: string): Promise<void> {
+    if (!window || !window.device) {
+      return Promise.reject(new Error("Registration requires cordova plugin. Verify the " +
+        "@aerogear/cordova-plugin-aerogear-push plugin is installed."));
+    }
+
+    if (!this.pushConfig || !this.pushConfig.config || !this.pushConfig.url) {
+      return Promise.reject(new Error("UPS registration: configuration is invalid"));
+    }
+
+    if (!deviceToken) {
+      return Promise.reject(new Error("Device token should not be empty"));
+    }
+
+    let platformConfig;
+    const url = this.pushConfig.url;
+    if (isCordovaAndroid()) {
+      platformConfig = this.pushConfig.config.android;
+    } else if (isCordovaIOS()) {
+      platformConfig = this.pushConfig.config.ios;
+    } else {
+      return Promise.reject(new Error("UPS registration: Platform is not supported."));
+    }
+
+    if (!platformConfig) {
+      return Promise.reject(new Error("UPS registration: Platform is configured." +
+        "Please add UPS variant and generate mobile - config.json again"));
+    }
+
+    const variantId = platformConfig.variantId || platformConfig.variantID;
+    if (!variantId) {
+      return Promise.reject(new Error("UPS registration: variantId is not defined."));
+    }
+
+    const variantSecret = platformConfig.variantSecret;
+    if (!variantSecret) {
+      return Promise.reject(new Error("UPS registration: variantSecret is not defined."));
+    }
+
+    this._objectInstance = window.PushNotification.init(
+      {
+        android: {},
+        ios: {
+          alert: true,
+          badge: true,
+          sound: true
+        }
+      }
+    );
+
+    const authToken = window.btoa(`${variantId}:${variantSecret}`);
+
+    const instance = axios.create({
+      baseURL: url,
+      timeout: 5000,
+      headers: {"Authorization": `Basic ${authToken}`}
+    });
+
+    return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


packages/push/src/PushRegistration.ts
```diff
@@ -96,22 +97,102 @@ export class PushRegistration {
...
+        }
+      )
+      .catch(reject);
+    });
+  }
+
+  /**
+   * Register deviceToken for Android or IOS platforms
+   */
+  public unregister(): Promise<void> {
```
wtrocki: How does this should work? Usually unregister is called when app is uninstalled but how do you plan to hook into the native callback. 

Unregister is usually done on native side.

danielpassos: @wtrocki Sorry, not sure if I understood your question here. Do you mean in which case/scenario the unregister should be used?

wtrocki: Is there any requirement for that or any use case that you see people to use it?
Do we plan to add that to showcase or explain how unregister should be used?

danielpassos: The requirement is that this feature is available on UPS from ages and also in all old SDKs. No, it was not added only because of the showcase switch [register/unregister]

Maybe _application will rarely have the switch like in showcase_ but I can give you a real scenario. After login, register the device after logout unregister the device but maybe the community has a bunch of real examples. That is a good thing to ask on ML.

When an app is deleted the device is not going to receive any messages anymore but the device will still "registered/listed" on UPS. The developer can try to unregister device when the app is deleted but unfortunately, this feature is only available on Android (as far as I know)

UPS will automatically purge the device from the database when Firebase return saying that is an invalid token and we also have a way to [purge an old token when the device registers the app again](https://github.com/aerogear/aerogear-unifiedpush-server/blob/ff5ba2cb8446d929e5fd426bcb645299b4559914/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/installations/InstallationRegistrationEndpoint.java#L132)

packages/push/src/PushRegistration.ts
```diff
@@ -96,22 +97,102 @@ export class PushRegistration {
...
+
+    const http = axios.create({
+      baseURL: url,
+      timeout: 5000,
+      headers: {"Authorization": `Basic ${authToken}`}
+    });
+
+    return new Promise((resolve, reject) => {
+      const endpoint = PushRegistration.API_PATH + "/" + deviceToken;
+      return http.delete(endpoint, {})
```
wtrocki: Missing catch

wtrocki:
Tried that. Looks great from showcase point of view but I did not actually hit the server and this actually crashed js thread because of missing catch method.

danielpassos:
@wtrocki Hrm, catch block is there on [line 199](https://github.com/aerogear/aerogear-js-sdk/blob/d6a0e573226090c7da37b22b0ebd4f2278a5168f/packages/push/src/PushRegistration.ts#L199)

packages/push/src/PushRegistration.ts
danielpassos: Hrm, there is weird. There is a [catch block there](https://github.com/aerogear/aerogear-js-sdk/blob/d6a0e573226090c7da37b22b0ebd4f2278a5168f/packages/push/src/PushRegistration.ts#L199).

wtrocki:
UPS server url for testing?

danielpassos:
Hrm, I'm running it locally. Sorry! Quickly way to run it:

1. Start postgres

   ```shell
   docker run \
       -e POSTGRES_USER=unifiedpush \
       -e POSTGRES_PASSWORD=unifiedpush \
       -e POSTGRES_DB=unifiedpush \
       -p 5432:5432 \
       postgres:11.2-alpine
   ```
1. Start the UPS

```shell
   docker run --rm \
       --name ups \
       -e POSTGRES_SERVICE_HOST=[YOUR_MACHINE_IP] \
       -e POSTGRES_SERVICE_PORT=5432 \
       -e POSTGRES_DATABASE=unifiedpush \
       -e POSTGRES_USER=unifiedpush \
       -e POSTGRES_PASSWORD=unifiedpush \
       -p 8080:8080 \
       aerogear/unifiedpush-wildfly-plain:2.2.1.Final
   ```

File: packages/push/src/PushRegistration.ts [hunk 3]
```diff
@@ -96,22 +97,102 @@ export class PushRegistration {
...
+
+    const storage = window.localStorage;
+    const deviceToken = storage.getItem(PushRegistration.TOKEN_KEY);
+
+    if (!deviceToken) {
+      return Promise.reject(new Error("Device token should not be empty"));
+    }
+
+    let platformConfig;
+    const url = this.pushConfig.url;
```
darahayes: Seems like a strange place to declare this. Would it not make sense to declare it down near line 185?

packages/push/src/PushRegistration.ts
```diff
@@ -96,22 +97,102 @@ export class PushRegistration {
...
+  /**
+   * Unregister device for Android or IOS platforms
+   */
+  public unregister(): Promise<void> {
+    if (!window || !window.device) {
+      return Promise.reject(new Error("Registration requires cordova plugin. Verify the " +
+        "@aerogear/cordova-plugin-aerogear-push plugin is installed."));
+    }
+
+    if (!this.pushConfig || !this.pushConfig.config || !this.pushConfig.url) {
```
darahayes: @danielpassos I see this code repeated here and in the `register` function. To me all this is doing is validating some configs that are actually passed into the constructor. Why not move all these checks there? Additionally, we're doing this `const http axios.create(...)` passing in some stuff that is all available from the configs passed into the constructor. I think it makes a lot of sense to do the following

* validate all of this config once in the constructor but do not throw (because I'm guessing we would break existing apps (showcase) where UPS is not always there). Set something like `this.validationError = <error message>`
* If no validation errors are present then set `this.httpClient = axios.create(...)` within the constructors
* then within `register` and `unregister` you can remove several lines of code and have a much more simple flow:

```
function register(...) : Promise<void> {
  if (this.validationError) {
    return Promise.reject(this.validationError)
  }

  ...
  this.httpClient.post(...) //or this.httpClient.delete()
}
```

That would really simplify this file and make it a lot more maintainable IMO. WDYT?

darahayes:
@danielpassos I left some feedback down there. Additionally, we should probably add a unit test for the `unregister` method.

danielpassos:
@darahayes Can you give a second eyes on this? 

I'm not sure what is the best way to raise the problems. On register/unregister return on reject:

1. Last problem found
1. List of the problems

packages/push/src/PushRegistration.ts
```diff
@@ -15,15 +19,85 @@ declare var window: any;
...
+      if (!this.validationError) {
+        const token = window.btoa(`${this.variantId}:${variantSecret}`);
+        this.http = axios.create({
+          baseURL: unifiedPushServerURL,
+          timeout: 5000,
+          headers: {"Authorization": `Basic ${token}`}
+        });
+      }
+
+      this.push = window.PushNotification.init(
```
darahayes: This I think this should probably stay in the `register` method right?

packages/push/src/PushRegistration.ts
```diff
@@ -15,15 +19,85 @@ declare var window: any;
...
-  public static readonly API_PATH: string = "rest/registry/device";
 
-  private readonly pushConfig?: ServiceConfiguration;
-  private _objectInstance: any;
+  private static readonly API_PATH: string = "rest/registry/device";
+  private static readonly TOKEN_KEY = "UPS_DEVICE_TOKEN";
+
+  private readonly validationError?: string;
+  private readonly variantId?: string;
+  private readonly http?: AxiosInstance;
```
darahayes: I think this should probably be something like `httpClient`

darahayes:
> I'm not sure what is the best way to raise the problems. On register/unregister return on reject:

> Last problem found
> List of the problems

@danielpassos I think that is entirely down to personal preference. I would probably just leave it as as. This would mean last problem found. Otherwise you are just building some internal array of errors and that is probably not necessary IMO.

packages/push/src/PushRegistration.ts
danielpassos: In the PR maybe but... I will need this in unregister to unsubscribe the device from the Firebase topics (next PR) so that is why I have move it to the constructor.

packages/push/src/PushRegistration.ts
danielpassos: Done!

packages/push/src/PushRegistration.ts
darahayes: I'm not sure what you mean. Do you mean you will need to access `this.push` from within the `unregister` method? That should be available because ideally the `register` method will have been called already before calling `unregister` right? And if you call `unregister` without having called `register` I would expect an error right?

From what I can see in the constructor, we are now guaranteed to always call `window.PushNotification.init()` even if `window.PushNotification` is undefined. This is a change in business logic because previously we only called it in the `register` method and only if `window.PushNotification` was defined. I think this could potentially cause some crashes if someone accidentally has a push config in their app while they run it in the browser.

If you think it's absolutely necessary to have it in the constructor it should at least be surrounded by the `if (!window || !window.device || !window.PushNotification)` statement on Line 37.

Also, we need unit tests.

danielpassos: > I'm not sure what you mean. Do you mean you will need to access this.push from within the unregister method? That should be available because ideally the register method will have been called already before calling unregister right? And if you call unregister without having called register I would expect an error right?

You can call register and than in another time need to call unregister but don't have the `PushRegistration` instance anymore, that is why I have added it in the construtor.

> From what I can see in the constructor, we are now guaranteed to always call window.PushNotification.init() even if window.PushNotification is undefined. This is a change in business logic because previously we only called it in the register method and only if window.PushNotification was defined. I think this could potentially cause some crashes if someone accidentally has a push config in their app while they run it in the browser.

Nice catch, I'm going to move it inside of `if (!this.validationError) {` block, because in that point we already had check `window.PushNotification`

PR #380, opened 2019-06-14
chore: use offix for cache functions

StephenCoady:
##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


wtrocki:
I used linked version to test with showcase. I did not tested restarts but it works like dream with regular edits.

wtrocki:
The fact that we decreased coverage by removing this means that coverage is probably missconfigured or something.

PR #381, opened 2019-06-14
added context based checking - not working yet will be rebased

StephenCoady:
### Description

@wtrocki I refactored this to take into account what we spoke about. its quick and dirty but the case we were testing earlier now works and is persisted to the server. So: 

1. Go offline client 1
1. Mutate title on client 1
1. Mutate description on client 2
1. Go online client 1
1. Happy scenario is now on server with title from client 1 and description from client 2

Outstanding to dos:

Need to save the base and returnType for offline mutations. They are being passed around in context but not persisted for now. Easy change.

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


wtrocki:
;) http://www.giphy.com/gifs/IcoIjcfSdNEN3u9TBc

PR #384, opened 2019-06-25
Unsubscribe Android device from Firebase topic when unregister from UPS

danielpassos:
# How to run UPS standalone

```
git clone https://github.com/aerogear/aerogear-unifiedpush-server.git
cd aerogear-unifiedpush-server
mvn clean install -DskipTests && docker run -p 9999:8080 -it aerogear/unifiedpush-configurable-container:2.2.2-SNAPSHOT
```

# How to test it

1. Register the device on UPS
1. Check if the device is listed on the UPS UI
1. Put the app in background
1. Send a message to the device (don't add any criteria) 
   * The device should receive a message show it on the 
1. Send a message to the device (using whatever category used) and check if the message was
   * The device should receive a message show it on the notification bar
1. Unregister the device on UPS
1. Check if the device gone from the UPS UI
1. Put the app in background
1. Send a message to the device (don't add any criteria) 
   * Make sure the device don't receive any message anymore
1. Send a message to the device (using whatever category used) and check if the message was
   * Make sure the device don't receive any message anymore

PS1: This step by step is by Android only
PS2: You can use [this branch](https://github.com/aerogear/ionic-showcase/tree/push-unsubscribe) to run this step by step

### DIFF
--- a/packages/push/src/PushRegistration.ts
+++ b/packages/push/src/PushRegistration.ts
@@ -106,53 +106,54 @@ export class PushRegistration {
   /**
    * Register deviceToken for Android or IOS platforms
    *
-   * @param deviceToken token that will be sent to Unified Push server
    * @param alias device alias used for registration
    * @param categories array list of categories that device should be register to.
    */
-  public register(deviceToken: string, alias: string = "", categories: string[] = []): Promise<void> {
-
-    if (!deviceToken) {
-      return Promise.reject(new Error("Device token should not be empty"));
-    }
+  public register(alias: string = "", categories: string[] = []): Promise<void> {
 
     if (this.validationError) {
       return Promise.reject(new Error(this.validationError));
     }
 
-    const postData = {
-      "deviceToken": deviceToken,
-      "deviceType": window.device.model,
-      "operatingSystem": window.device.platform,
-      "osVersion": window.device.version,
-      "alias": alias,
-      "categories": categories
-    };
-
     return new Promise((resolve, reject) => {
-      if (this.httpClient) {
-        return this.httpClient.post(PushRegistration.API_PATH, postData)
-        .then(
-          () => {
 
-            if (isCordovaAndroid() && this.variantId) {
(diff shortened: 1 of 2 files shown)


packages/push/src/PushRegistration.ts
```diff
@@ -106,52 +106,54 @@ export class PushRegistration {
...
-          () => {
-            const storage = window.localStorage;
-            storage.setItem(PushRegistration.TOKEN_KEY, deviceToken);
-
-            if (isCordovaAndroid() && this.variantId) {
-              this.subscribeToFirebaseTopic(this.variantId);
-              for (const category of categories) {
-                this.subscribeToFirebaseTopic(category);
+
+      setTimeout(() => reject(), 5000);
```
darahayes: Should probably include some message in this rejection

danielpassos:
Yes. iOS will be able to receive notification also. This change is only for Android because only Android uses [Firebase topics|https://firebase.google.com/docs/cloud-messaging/android/topic-messaging]. We don't have that implemented to iOS yet.

darahayes:
@danielpassos so what will the flow look like for iOS users will they still have to perform their own registration step?

packages/push/src/PushRegistration.ts
danielpassos: Done. I will also provides a way to developer configure the timeout but I have it in place on the next PR ;)

danielpassos:
@darahayes Nopz. Registration processe will be handle from the js-sdk [over here|https://github.com/aerogear/aerogear-js-sdk/blob/57fd2de588fba7b9908fa0fb9038a7d3060715f8/packages/push/src/PushRegistration.ts#L122] for both (Android and iOS)

danielpassos:
Just to be more clean the only difference between Android and iOS is Android need to register the device on the Firebase topics because when UPS wanna send a message for a group of devices UPS uses topics for Android, so it sends only one message for all group of devices and on iOS land UPS fetch all devices in the UPS database and send 1 message for each device.

PR #385, opened 2019-06-27
fix: removed extra `/` from URL

craicoverflow:
### Description

https://issues.jboss.org/browse/AEROGEAR-9496

Remove an extra forward slash from the `/init` endpoint. The route generated by Ingress already includes a trailing forward slash so this needs to be removed.

Example Ingress route: `http://route-mobile-security-service.192.168.42.171.nip.io/`

In its present state, the init endpoint would be:

`http://route-mobile-security-service.192.168.42.171.nip.io//api/init`

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


wtrocki:
@craicoverflow Minor mock needs to be adjusted in tests. Approving

PR #386, opened 2019-06-27
fix: removed extra `/` from URL

craicoverflow:
### Description

https://issues.jboss.org/browse/AEROGEAR-9496

Remove an extra forward slash from the `/init` endpoint. The route generated by Ingress already includes a trailing forward slash so this needs to be removed.

Example Ingress route: `http://route-mobile-security-service.192.168.42.171.nip.io/`

In its present state, the init endpoint would be:

`http://route-mobile-security-service.192.168.42.171.nip.io//api/init`

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### How

Used [The WHATWG URL API](https://nodejs.org/api/url.html#url_the_whatwg_url_api).

### Additional Notes

Tested with the Mobile Security Service and ionic-showcase:

Screenshot from 2019-06-27 14-02-42

### DIFF
--- a/packages/security/src/appSecurity/AppSecurity.ts
+++ b/packages/security/src/appSecurity/AppSecurity.ts
@@ -70,9 +70,9 @@ export class AppSecurity {
    *
    * @param baseUrl
    */
-  getInitUrl(baseUrl: string): string {
+  private getInitUrl(baseUrl: string): string {
     const url = new URL(baseUrl);
-    url.pathname = '/api/init';
+    url.pathname = "/api/init";
     return url.href;
   }
 }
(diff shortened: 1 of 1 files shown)


packages/security/src/appSecurity/AppSecurity.ts
```diff
@@ -66,4 +64,15 @@ export class AppSecurity {
...
   }
+
+  /**
+   * Build the init URL from the base URL
+   *
+   * @param baseUrl
+   */
+  private getInitUrl(baseUrl: string): string {
+    const url = new URL(baseUrl);
+    url.pathname = "/api/init";
```
austincunningham: is this adding the slash again

craicoverflow: No. `pathname` replaces the existing forward slash.

https://nodejs.org/api/url.html#url_url_pathname

craicoverflow:
Closing this as it is not needed - fixed client side instead.

PR #388, opened 2019-07-09
Use released conflict implementation 

wtrocki:
WIP


darahayes:
Are those integration tests moved over to offix or somewhere else?

wtrocki:
Yes. Integration tests will be maintained in offix as this is where all of the functionalities are and we will have downstream integration tests covering all products including sync

PR #393, opened 2019-07-10
fix: remove dev dependency types

wtrocki:
##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


wtrocki:
Problem is with offix packages. Resolving it there.

PR #394, opened 2019-07-10
RELEASE_NOTES.md

wtrocki:
Idea for this file will be to stage release notes in the repository that will help me with releases. Our release notes are very thin now this file will allow us to drop more things into the release notes.


wtrocki:
@danielpassos Once this is merged. Can you drop content from the email to this file. 
We need to start pushing more info in releases. THX

RELEASE_NOTES.md
```diff
@@ -0,0 +1,3 @@
+## Release Notes
```
danielpassos: It should be `#` instead of `##`

PR #402, opened 2019-07-12
chore: added script to make development easier

StephenCoady:
### Description

added a small script to the base of the sdk to make developing and testing the sdk much easier.

NOTE: it assumes offix is in the same directory as the aerogear-sdk

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
     "update:check": "lerna exec --concurrency 1 -- ncu",
     "update:run": "lerna exec -- ncu -a --removeRange -x 'fh-js-sdk'",
     "generate:docs": "./scripts/docs.sh",
-    "version:master": "./scripts/versionMaster.sh"
+    "version:master": "./scripts/versionMaster.sh",
+    "dev": "./scripts/syncDevEnvironment.sh"
   },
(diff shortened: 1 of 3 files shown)


wtrocki:
Awesome work!

darahayes:
Really nice PR, a small suggestion would be to document the fact that it exists and what it does somewhere. Perhaps the Readme?

wtrocki:
I guess the same will happen for the voyager-server?

wtrocki:
For actual linking problem.
I actually started to hack cordova by specifying `alias npm=yarn` and it does the job as install command is the same, but this works only when we link things using yarn rather than NPM. Yarn links are better

wtrocki:
@StephenCoady I will be happy to merge this and we can experiment with alias on yarn on our own. I alised that globally and going to see if I will see any issues.

wtrocki:
@jhellar Do you think we can have some label that circleCI will read to skip master release. There will be actually tones of changes that are not code ones and we can save a lots of traffic and time if we skip the steps.

PR #404, opened 2019-07-12
Release master

jhellar:
### Description

Motivation: https://issues.jboss.org/browse/AEROGEAR-9334

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,7 +73,7 @@ jobs:
       - checkout
       - run:
           name: version packages
-          command: npx lerna version prerelease --no-push --yes --preid "master.$(git rev-list HEAD --count).$(git rev-parse --short HEAD)"
+          command: npm run version:master
       - publish:
           tag: "$(node -e \"console.log(require('./lerna.json').version);\")"
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,10 +1,21 @@
-version: 2
+version: 2.1
+commands:
+  publish:
+    parameters:
+      tag:
+        type: string
+    steps:
+        # Allows us to authenticate with the npm registry
+      - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/.npmrc
+      - run: CI=true npm run release:prep
+      - run: TAG=<< parameters.tag >> npm run release:validate
+      - run: TAG=<< parameters.tag >> npm run publish
+
 jobs:
   build:
     working_directory: ~/aerogear
     docker:
-      # Node 8 LTS
-      - image: circleci/node:lts-browsers
+      - image: circleci/node:lts
     steps:
       - checkout
       - run:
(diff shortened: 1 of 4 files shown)


scripts/publishRelease.sh
```diff
@@ -4,6 +4,7 @@
 TAG=$TAG
 
 RELEASE_SYNTAX='^[0-9]+\.[0-9]+\.[0-9]+$'
+MASTER_PRERELEASE_SYNTAX='^[0-9]+\.[0-9]+\.[0-9]+-master\..+$'
```
wtrocki: This is kinda pointless as we can just simply check if tag equals master?

.circleci/config.yml
```diff
@@ -1,10 +1,21 @@
-version: 2
+version: 2.1
+commands:
+  publish:
+    parameters:
+      tag:
+        type: string
+    steps:
+        # Allows us to authenticate with the npm registry
+      - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/.npmrc
```
wtrocki: We should use scripts for everything that is more than single line

.circleci/config.yml
```diff
@@ -49,15 +60,22 @@ jobs:
...
+          tag: "$CIRCLE_TAG"
+  npm_publish_master:
+    working_directory: ~/aerogear
+    docker:
+      - image: circleci/node:lts
+    steps:
+      - checkout
+      - run:
+          name: version packages
+          command: npx lerna version prerelease --no-push --yes --preid "master.$(git rev-list HEAD --count).$(git rev-parse --short HEAD)"
```
wtrocki: I would really want to have script that does this stuff so we can try this before merge on local machine and have way to replicate this changes across repositories (voyager, js and others)

scripts/publishRelease.sh
jhellar: the tag will be actually something like `2.6.0-master.967.57a7482`, then dist-tag `master` should point to it

.circleci/config.yml
darahayes: Each of these individual steps is a script. Wrapping those ones up into a bigger script obfuscates how the release process works. I think this usage of CircleCI commands to not duplicate the npm publishing steps is really nice stuff.

jhellar:
Thanks for review. With more master versions on npm I agree that this might not be a good idea. Would it make sense to keep only the latest master on npm - with every master release unpublish the previous one? @darahayes @wtrocki

wtrocki:
I think from my side suggestion will be to use `dev` instead of `master` tag as we already have dev and the purpose of those is exactly the same. Apart from that I'm happy to see how this will play out. 

A side impact of this change is that I will probably need to do less renovate merges. Aggregate them somehow. Renovate tends to create so many PR's and it can mean 5 releases in an hour where we address them. Obviously, we can wait with addressing those and this is just a workflow change.

jhellar:
@wtrocki I have changed it so that `dev` tag is used.

scripts/publishRelease.sh
darahayes: @jhellar now that you're using the `dev` tag on npm. Maybe we should change this to use the word `dev` instead of master also?

jhellar: @darahayes you mean just for the version/variable name, or should it be changed everywhere - not mention master but dev?

darahayes: Yeah I mean the actual value so you will have something like `2.6.0-dev.967.57a7482`

jhellar: ok, makes sense, will change it

wtrocki:
I will ping you on next merge so we can verify that in action

PR #406, opened 2019-07-17
fix: remove clientId from sync header payloads

darahayes:
### Description

Removes the `clientId` property from the header that would be sent by the sync SDK as part of websocket connections. Relates to https://github.com/aerogear/keycloak-connect-graphql/pull/23

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -129,8 +129,7 @@ export class Auth {
               header: {
                 "Authorization": "Bearer " + this.auth.token
               },
-              token: this.auth.token,
-              clientId: this.internalConfig.clientId
+              token: this.auth.token
             });
           } else {
             reject("No keycloak token available");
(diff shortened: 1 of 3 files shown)


wtrocki:
Change to `BREAKING:`. Add note to changelog so we can point people when they get issues on showcase on their sample apps.

wtrocki:
I could not find any documentation change, but I will recommend to review if we mention token in mobile docs.

darahayes:
@wtrocki we don't. In fact the docs in this particular area are lacking a lot

wtrocki:
Requires a change in upstream as well: https://github.com/aerogear/offix/blob/master/packages/offix-client/src/auth/AuthContextProvider.ts

darahayes:
> Requires a change in upstream as well

@wtrocki we're changing what we have here to be the same as what is in the upstream package so no change is required.

PR #408, opened 2019-07-17
configure git user

jhellar:
### Description

Publishing master fails now [1]. When lerna versions packages, it needs to commit the changes. For this git user needs to be configured.

[1] https://circleci.com/gh/aerogear/aerogear-js-sdk/2068?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,6 +72,12 @@ jobs:
     steps:
       - checkout
       - run:
+          name: configure git user.email
+          command: git config --global user.email "test@example.com"
+      - run:
+          name: configure git user.name
+          command: git config --global user.name "test"
+      - run:
           name: version packages
           command: npm run version:master
       - publish:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,12 +72,6 @@ jobs:
     steps:
       - checkout
       - run:
-          name: configure git user.email
-          command: git config --global user.email "test@example.com"
-      - run:
-          name: configure git user.name
-          command: git config --global user.name "test"
-      - run:
           name: version packages
           command: npm run version:master
       - publish:
--- a/scripts/versionMaster.sh
+++ b/scripts/versionMaster.sh
@@ -4,4 +4,4 @@ NUM_COMMITS="$(git rev-list HEAD --count)"
 LAST_COMMIT="$(git rev-parse --short HEAD)"
 PRE_ID="dev.$NUM_COMMITS.$LAST_COMMIT"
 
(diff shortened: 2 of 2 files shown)


darahayes:
-1 on this approach. Lerna will try to commit and push those changes back which we definitely do not want. My bad that I forgot about this when originally reviewing the PR that introduced this script.

Lerna has flags to disable that behaviour. `--no-git-tag-version` and `--no-push` should be used instead. https://github.com/lerna/lerna/tree/master/commands/version#--no-git-tag-version

jhellar:
It actually uses `--no-push` flag not to push back to remote repo. But it still needs git user to commit changes locally. Didn't know about `--no-git-tag-version` flag, will give it a try.

jhellar:
@darahayes tried also `--no-git-tag-version`, lerna still commits the changes, just does not create a tag.

wtrocki:
Solid -1 on this. I will need some time to check how we can do that effectively. 
We agreed to have some additional artifacts on npm, but having this in repository is too much. 

EDIT: 

I have put some thought and checked how this problem is resolved in wider open-source land and people build release bots. I really love idea to have boot that will


Create Release PR. Rebase changes during the day. Automerge the changes by the end of the day and create new release. 

This way we not only can use this for master releases but also do aerogear releases as whole. Codebase for release bot can be reused - there are couple open source librarires for doing bots. This way we do not do any shady stuff like circleci commiting the code - this sounds really really scary :)
 

WDYT?

darahayes:
@jhellar that might be worth opening an issue in Lerna. In any case, the following command is used in `release:prep` which does exactly what you want. https://github.com/aerogear/aerogear-js-sdk/blob/master/scripts/prepareRelease.sh#L15

It's kind of weird and it uses deprecated arguments but it works. It might be worth investigating to see can you replicate that behaviour using non deprecated flags

jhellar:
@wtrocki there would not be any changes in repository. When lerna bumps versions of packages, it commits the changes (locally), but it won’t push them to GitHub.

The idea with bot sounds also good.

@darahayes tried the `--skip-git`, works fine - it won’t commit the changes, just the deprecation message.

jhellar:
Tried the combination of `--no-git-tag-version` and `--no-push` - that also won't commit the changes - will update this PR.

jhellar:
@wtrocki @darahayes would this fix be sufficient for now?

wtrocki:
@jhellar Yes.. Sorry for suggesting a lot of work - that will be cool to have.

PR #410, opened 2019-07-19
fix: version validation

jhellar:
### Description

publishing master fails: https://circleci.com/gh/aerogear/aerogear-js-sdk/2076?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

This fixes script that validates versions.

### DIFF
--- a/scripts/validateRelease.sh
+++ b/scripts/validateRelease.sh
@@ -24,7 +24,7 @@ fi
 
 # validate that all packages have the same version found in lerna.json
 for package in $PACKAGES; do
-  version=$(lerna --loglevel=silent ls -l | grep $package | awk -F ' ' '{print $2}' | awk -F 'v' '{print $2}')
+  version=$(lerna --loglevel=silent ls -l | grep $package | awk -F ' ' '{print $2}' | cut -c2-)
   if [[ $version =~ $PACKAGE_VERSION ]]; then
     echo "package $package has version $version"
   else


wtrocki:
For maintenance over the future we can simply use node.js codebase instead of bash so it will be easier to track issues and test it.

jhellar:
@darahayes I came up with something like this `version=$(package=$package node -e 'console.log(require(\'./packages/\' + process.env.package.replace(\'@aerogear/\', \'\') + \'/package.json\').version)')` - which looks quite ugly, the packages are not yet published&installed, so I was not able to use `require(process.env.package)`.

@wtrocki do you mean rewriting all the bash scripts to js? would it be ok to create just github issue for it for now?

wtrocki:
@jhellar yes. Do not get that we need to rewrite those, but we can practically have some simple mixed bash/node scripts for certain tasks. This is executed in node.js environment so we can assume that node is installed. It will be also easy to reuse this as we can publish all release scripts as package. Now I'm replicating those across 9 different repositories. Any custom release cases can be nicely coded in node.

darahayes:
> For maintenance over the future we can simply use node.js codebase instead of bash so it will be easier to track issues and test it.

> It will be also easy to reuse this as we can publish all release scripts as package.

On these two points: I am +10000 on solving the problem we have with duplication of release/build/deployment tooling across repos. However I think, let's not jump to the immediate conclusion that the solution is a package. I think this is something we could spike or at least have a chat about to explore our options.

wtrocki:
I meant to have helpers package for validating versions and all other stuff that is hard in bash. We can still have bash scripts for the rest

PR #411, opened 2019-07-19
breaking: refactor AuthContextProvider interfaces, remove token

darahayes:
### Description

For verification, if you link up these changes into the showcase app, you should be able to follow the same steps provided here: https://github.com/aerogear/ionic-showcase/pull/228

And ideally you shouldn't see any issues.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -126,10 +126,9 @@ export class Auth {
         tokenUpdate.success(() => {
           if (this.auth.token) {
             resolve({
-              header: {
+              headers: {
                 "Authorization": "Bearer " + this.auth.token
-              },
-              token: this.auth.token
+              }
             });
           } else {
             reject("No keycloak token available");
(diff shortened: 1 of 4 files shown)


wtrocki:
I would like to extend exsisting unit test or integrationm test with auth context provider so we can kinda fail if things go bad. 

With current form this is more like header provider than auth context provider as it is not auth specific. People can still drop their link to apply any custom headers so we are ok.

wtrocki:
I would like to extend the exsisting unit test or integration test with the auth context provider so we can kinda fail if things go bad. 

With current form, this is more like header provider than auth context provider as it is not auth specific. People can still drop their link to apply any custom headers so we are ok.

darahayes:
> With current form, this is more like header provider than auth context provider as it is not auth specific. People can still drop their link to apply any custom headers so we are ok.

I totally agree. It makes things more flexible but still allows us to provide what we need in terms of keycloak -> sync integration

wtrocki:
@darahayes Can you also add info to RELEASE_NOTES.md

darahayes:
PR for release notes: https://github.com/aerogear/aerogear-js-sdk/pull/412

PR #412, opened 2019-07-19
chore: update release notes

darahayes:
### Description

@wtrocki @StephenCoady I've added an **Unreleased** section following a format/standard called [Keep a Changelog](https://keepachangelog.com/en/0.3.0/)

When we go to cut a release we simply rename the unreleased heading to the new version and start a new unreleased section. Let me know what you think.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,3 @@
-# Release Notes 
-
 ## JS-SDK 2.6.0
 
 ### DataSync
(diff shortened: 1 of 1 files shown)


wtrocki:
Like the changes but we will need to apply this to entire 2.6 before merge.

wtrocki:
@darahayes Can you reformat this notes as we doing release now?

wtrocki:
We can use different format from upcomming release

darahayes:
@wtrocki sure. I'll follow the same format you guys are using for 0.7.0 right?

wtrocki:
Yes. We so we might reformat it or join both. I do not mind but we need that to be put release notes with 2.6.0

PR #424, opened 2019-08-08
fix: integrate with datasync

wtrocki:
WDYT about this? It requires release of offix


packages/sync/src/DataSyncClient.ts
```diff
@@ -0,0 +1,49 @@
...
+  constructor(private userConfig: DataSyncConfig) {
+    super(userConfig);
+  }
+
+  /**
+   * Initialize client
+   */
+  public async init(): Promise<ApolloOfflineClient> {
+    const terminatingLink = await createDefaultLink(this.userConfig);
+    this.config.terminatingLink = terminatingLink;
```
wtrocki: That was the reason I wanted init method to get terminatingLink but we can expose config and do it as well.

darahayes: I mean, it would be possible to simply call `createDefaultLink` within `createClient` and add it to the config. Then you wouldn't need to override `init` at all right?

wtrocki: Do you think we can return ApolloClient - do you need OfflineClient for anything in your current use cases?

wtrocki:
@darahayes Requires package update

wtrocki:
Ready to merge from my side!

PR #431, opened 2019-09-18
Added web-push support

ziccardi:
# Motivation
**JIRA**: https://issues.jboss.org/browse/AEROGEAR-9895

### Description

- Added support for the webpush
- Changed the library so that it can used without `cordova` (for webpush)
- Refactored the code adding some OO structure

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/push/src/PushRegistration.ts
+++ b/packages/push/src/PushRegistration.ts
@@ -2,7 +2,6 @@ import {
   ConfigurationService,
   isNative
 } from "@aerogear/core";
-import { AbstractPushRegistration } from "./AbstractPushRegistration";
 import {
   PushRegistrationInterface,
   PushRegistrationOptions,
--- a/packages/push/src/PushRegistration.ts
+++ b/packages/push/src/PushRegistration.ts
@@ -24,6 +24,11 @@ export interface PushRegistrationOptions {
  * registration.register("myAppleOrFirebaseToken");
  */
 export class PushRegistration implements PushRegistrationInterface {
+
+  public static onMessageReceived(onMessageReceivedCallback: OnMessageReceivedCallback) {
+    PushRegistrationCordovaImpl.onMessageReceived(onMessageReceivedCallback);
+  }
+
   private readonly delegate: AbstractPushRegistration;
 
   constructor(config: ConfigurationService) {
--- a/packages/push/src/PushRegistration.ts
+++ b/packages/push/src/PushRegistration.ts
@@ -18,7 +18,6 @@ export class PushRegistration implements PushRegistrationInterface {
   private readonly impl: AbstractPushRegistration;
 
   constructor(config: ConfigurationService) {
-    console.warn('constructor');
     if (!isNative()) {
       // impl must be a non cordova impl
       this.impl = new PushRegistrationWebpushImpl(config);
--- a/packages/push/src/PushRegistration.ts
+++ b/packages/push/src/PushRegistration.ts
@@ -14,26 +14,38 @@ export interface PushRegistrationOptions {
   timeout?: number;
 }
 
+/**
+ * AeroGear UPS registration SDK
+ *
+ * Usage:
+ * // Initialize SDK first
+ * app.init(config);
+ * let registration = new PushRegistration();
+ * registration.register("myAppleOrFirebaseToken");
+ */
 export class PushRegistration implements PushRegistrationInterface {
-  private readonly impl: AbstractPushRegistration;
+  private readonly delegate: AbstractPushRegistration;
 
   constructor(config: ConfigurationService) {
     if (!isNative()) {
-      // impl must be a non cordova impl
-      this.impl = new PushRegistrationWebpushImpl(config);
+      this.delegate = new PushRegistrationWebpushImpl(config);
     } else {
-      this.impl = new PushRegistrationCordovaImpl(config);
+      this.delegate = new PushRegistrationCordovaImpl(config);
     }
   }
 
+  /**
+   * Register the application to the UPS.
+   * @param options
+   */
   public async register(options: PushRegistrationOptions = {}): Promise<void> {
-    await this.impl.register(options);
+    await this.delegate.register(options);
   }
+
+  /**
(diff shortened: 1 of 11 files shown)


wtrocki:
@ziccardi Initial idea of the package was that it will handle registration only and it could be used in cordova and other plugins. Then we kinda brought some Cordova specific code here, but we might revise this completely and move out cordova and have a separate packages for webpush. 

Generally, all depends on how much code/changes/dependencies we will need to add here.
At this point, it looks like the amount of the changes will be small and we are ok with this.

packages/push/src/PushRegistrationInterface.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export interface PushRegistrationWebpushOptions extends PushRegistrationOptions {
+  serviceWorker?: string;
+}
+
+/**
+ * Interface for PushRegistration objects.
+ */
+export interface PushRegistrationInterface {
```
danielpassos: I know what you did here, but to be very honest for this case I think we don't need a `PushRegistrationInterface`

packages/push/src/PushRegistration.ts
```diff
@@ -24,249 +22,38 @@ export interface PushRegistrationOptions {
...
-          resolve();
-        })
-        .catch(reject);
-      } else {
-        // It should never happend but...
-        return reject(new Error("Push is not properly configured"));
-      }
-    });
+  public async unregister(): Promise<void> {
+    this.delegate.unregister();
```
danielpassos: Do we need to add an `await` here like in `register()`?

packages/push/src/PushRegistration.ts
```diff
@@ -24,249 +22,38 @@ export interface PushRegistrationOptions {
...
-  public static readonly REGISTRATION_DATA_KEY = "UPS_REGISTRATION_DATA";
+export class PushRegistration implements PushRegistrationInterface {
 
+  /**
+   * Calls to this method are ignored if webpush are used.
+   * @param onMessageReceivedCallback
+   */
   public static onMessageReceived(onMessageReceivedCallback: OnMessageReceivedCallback) {
-    PushRegistration.onMessageReceivedCallback = onMessageReceivedCallback;
+    PushRegistrationCordovaImpl.onMessageReceived(onMessageReceivedCallback);
```
danielpassos: We should delegate it. Also It's not going to work for WebPush

packages/push/src/impl/PushRegistrationWebpushImpl.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    if (this.validationError) {
+      throw new Error(this.validationError);
+    }
+
+    if (this.httpClient) {
+      const subscription = await this.subscribeToWebPush(this.platformConfig.appServerKey, serviceWorker);
+      const postData = {
+        "deviceToken": JSON.stringify(subscription),
+        "deviceType": "ChromeBrowser",
```
danielpassos: Maybe we should think in a way to get the browser name

packages/push/src/impl/PushRegistrationWebpushImpl.ts
```diff
@@ -0,0 +1,141 @@
...
+        "operatingSystem": "WebPush",
+        "osVersion": "6.1.2",
```
danielpassos: Maybw we should think in a way to get the browser version

packages/push/src/impl/PushRegistrationWebpushImpl.ts
```diff
@@ -0,0 +1,141 @@
...
+      await this.httpClient.post(AbstractPushRegistration.API_PATH, postData);
+      const storage = window.localStorage;
+      storage.setItem(AbstractPushRegistration.REGISTRATION_DATA_KEY, JSON.stringify(postData));
+    } else {
+      // It should never happend but...
+      throw new Error("Push is not properly configured");
+    }
+  }
+
+  // tslint:disable-next-line: no-empty
```
danielpassos: Remove?

packages/push/src/PushRegistration.ts
ziccardi: @danielpassos It delegates it to the `PushRegistrationCordovaImpl`. In the comment it is documented that calls to this method are ignored for WebPush.

The whole goal of using delegates here is to make the outer interface unchanged and, unfortunately, the old interface had this static method, so I had to leave it there to keep the interface unchanged.

packages/push/src/PushRegistration.ts
ziccardi: 👍

packages/push/src/PushRegistration.ts
ziccardi: I added a `console.warn` message that is visualised if calling that method while using webpush

packages/push/src/impl/PushRegistrationWebpushImpl.ts
secondsun: There's `window.navigator.userAgent` which may be useful.

secondsun:
Just a quick note, 

Chrome and firefox both work great on Windows with this.

Edge can register just fine, but unregistration causes a problem due to the "?" token in the endpoint field in the registration info.  We use this registration info in place of the device token.

Perhaps we should discuss transforming the webpush registration information into some way (base64 encode?) and change how the server treats the it.

packages/push/src/PushRegistration.ts
ziccardi: No need to have that comment anymore: with the latest changes it is now used for webpush too.

ziccardi:
> Just a quick note,
> 
> Chrome and firefox both work great on Windows with this.
> 
> Edge can register just fine, but unregistration causes a problem due to the "?" token in the endpoint field in the registration info. We use this registration info in place of the device token.
> 
> Perhaps we should discuss transforming the webpush registration information into some way (base64 encode?) and change how the server treats the it.

Done!

PR #434, opened 2019-10-02
Release 2.7.0

danielpassos:


RELEASE_NOTES.md
```diff
@@ -1,3 +1,9 @@
+## JS-SDK 2.7.0
+
+### Push
+
+Support WebPush
```
wtrocki: Can this be extended little bit.

danielpassos: What kind of info should I add?

secondsun: How about 
#### WebPush Support
 * WebPush Variants in mobile-services.json are now supported
 * Compatible browsers such as Chrome can register for and receive webpush push messages sent from UPS
 * Framework for connecting custom service-workers to notification APIs
 * Permissions for Notifications are handled by JS-SDK

danielpassos: Done! @wtrocki Do you think we need some else info?

wtrocki:
Works for me now!

PR #443, opened 2019-11-19
BREAKING: remove not used packages and MDC based config

wtrocki:
Showcase PR: https://github.com/aerogear/ionic-showcase/pull/277

### Removing security and metrics packages

Both security and metrics were removed. Leaving auth as it has some useful method that we do not want to port so fast.

### Config that works for developers

We removing need for the config - with versions etc. towards passing individual service configs to the services. 

This change is breaking previous behavior, however it will simplify documentation efforts as we going to pass only service configuration rather than the entire object that would make no sense for individual service:

Screenshot 2019-11-20 at 10 01 54

## Notes

This represents version JS SDK 3.0 I always wanted to draft that utilizes TypeScript generics for configs - so documentation is not that important if you are getting compilation errors.

I still left URL on top level for simplicity - like datasync config that requires technically only url and more specialized config. This is easy to document and it will be typechecked for each SDK.

- For Auth - uses any as I could not find any config typing. Maybe we should write our own???
- For Sync - business as usual - created typings for config - it will simplify the showcase app as well.
- For Push - Typings will be needed - looking for guidance/example

Core package and auth can easily go away, but I do not want to cause extra hassle on the showcase app level. This changes can be ported in showcase in just couple minutes. 

## Backwards compatibility

Backwards compatibility for this changes is very simple. 
Developers can still use mobile-services.json however they will need to manually pass individual service config:
```
new Auth(config.services.find(obj => obj.name === 'keycloak')
```

However, for the new usages, it will be much easier to work with when copying the config from keycloak

```
new Auth({config})
```

I somehow question now value of keycloak SDK (it makes no sense to have it since there is simply nothing we offer there.

## What is outstanding

- [ ] Improve typings
- [ ] Adjust documentation
- [ ] Build release notes for 3.0 (new and breaking channges from offix will need to land here as well)


wtrocki:
@darahayes @secondsun FYI

wtrocki:
Backwards compatibility for this changes is very simple. 
Developers can still use mobile-services.json however they will need to manually pass individual service config:
```
new Auth(config.services.find(obj => obj.name === 'keycloak')
```

However, for the new usages, it will be much easier to work with when copying the config from keycloak

```
new Auth({config})
```

I somehow question now value of keycloak SDK (it makes no sense to have it since there is simply nothing we offer there./

secondsun:
I'm riffing on push config here for a second without coffee, so forgive nonsense :)

First I'm "meh" with `type` , `url` and `config` as top level properties.  I don't think they are that helpful and in the case of push-config.json the url property is duplicated by the pushServerURL.  Additionally the whole file can be a stand alone config.  If they are important for backwards compatibility they are fine, but we might want a way to work around them.

Second, here's an example push config : https://gist.github.com/secondsun/0e6fa612640c5f8f119dd7070bfb42dd

the format is 
```{
  "pushServerURL":"http://some_url",
  "$Variant_type": {
        "senderID": "$variant_name", #only in android variant types,
        "appServerKey": "A public key", #only in web_push variant types
        "variantID": "$UUID",
        "variantSecret": "$UUID"
  }
}```

`$Variant_type` can be one of `android`, `web_push`, or `ios`.  

There are a couple of things here that we can do to make this better; senderid isn't required in Android since firebase has been a thing.  We also are shipping UPS with webpush so the format could be simplified to just `variantId`, `variantSecret`, `url`, and `variant type`, but that's work for the future perhaps.

**Comment by wtrocki** [comment]
> If they are important for backwards compatibility they are fine, but we might want a way to work around them.

Actually they are not. I just kept URL because I was afraid of negative feedback, but if we are pro dropping entire url thing that works even better. 

@darahayes Any opinions for offix?

**Comment by wtrocki** [comment]
How those typings will work for you

```ts
interface PushVariant{
  senderID?: string,
  appServerKey: string,
  variantID: string
  variantSecret: string
}

interface PushConfig {
  pushServerURL: string,
  android?: PushVariant,
  ios?: PushVariant,
  web?: PushVariant
}
```

API will look like this:
```
 config: PushConfig = ... 
 push = new PushRegistration(config)
```

darahayes:
> @darahayes Any opinions for offix?

So there is the actual static config that used to come from the MDC which was just a `url` and a `wsUrl`.

But then we have the actual `DataSyncConfig` which includes lots of extra flags and listeners and so on.

IMO there are two options and I don't mind which one we do.

We could have current `DataSyncConfig` + `SyncServiceConfig`

```ts
interface SyncServiceConfig {
  url: string
  wsUrl: string
}
```

and the API will look something like this:

```ts
//
const client = await createClient(serviceConfig)

// or when we have extra modifiers for the client
const client = await createClient(serviceConfig, { fileUpload: true, queueListeners: {...}, ...}
```

Or we can put them together like


```ts
interface SyncServiceConfig {
  url: string
  wsUrl: string
}

interface SyncClientConfig extends SyncServiceConfig {
  storage?: PersistentStore<PersistedData>;
  networkStatus?: NetworkStatus;
  offlineQueueListener?: OfflineQueueListener<MutationOptions>;
  authContextProvider?: AuthContextProvider;
  auditLogging?: boolean;
  fileUpload?: boolean;
  conflictProvider?: ObjectState;
  conflictListener?: ConflictListener;
  conflictStrategy?: ConflictResolutionStrategy;
  mutationCacheUpdates?: CacheUpdates;
  retryOptions?: RetryLink.Options;
}
```

and API will be like:

```ts
const client = await createClient(serviceConfig)

// and when extra modifiers are needed
const client = await createClient({
  ...serviceConfig
  networkStatus: {...},
  mutationCacheUpdates: {...},
})

I'm happy with either but I'd be leaning towards the second one.

wtrocki:
Second option is how I actually have implemented that in showcase:
https://github.com/aerogear/ionic-showcase/blob/cfaf8ea5edb5b5393d040594e1718b16c2205b7d/src/app/services/sync/voyager.service.ts#L81-L86

So if we go with the second option on api level then we have no difference betewen configs, but it will be nicer to recognize this 2 flags as modifiers :)

wtrocki:
Closing as we moving to deprecate js-sdk

